PR #5, opened 2018-10-02
Map component view style

kosmidma:


angular.json
```diff
@@ -28,7 +28,8 @@
             ],
             "styles": [
               "node_modules/ol/ol.css",
-              "src/scss/styles.scss"
+              "src/scss/styles.scss",
+              "css/all.css"
```
jswk: The path should read "node_modules/@fortawesome/fontawesome-free/css/all.css". Also, as it is a library you import it should go before "src/scss/styles.scss".

sat4envi.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jswk: This file shouldn't be commited. Add a line "*.iml" to /.gitignore to avoid commiting it and remove it from the repo in a next commit.

src/assets/i18n/pl.json
```diff
@@ -8,6 +8,6 @@
   "s4e-view-manager": {
     "OVERLAYS": "Nakładki",
     "PRODUCTS": "Produkty",
-    "REMOVE_PRODUCT_VIEW": "Usuń"
+    "REMOVE_PRODUCT_VIEW": ""
```
jswk: This line should be removed completely, not just set to an empty string

src/scss/bootstrap-view.scss
```diff
@@ -0,0 +1,94 @@
...
+  font-size: 14px;
+  line-height: 130%;
+}
+
+.border-left {
+  border-left: 1px solid $gray-800!important;
+}
+
+a {
+  color: $gray-900!important;
```
jswk: Don't use !important on a global rule.

src/scss/bootstrap-view.scss
```diff
@@ -0,0 +1,94 @@
...
+h2{
+  background-color: $gray-200;
+  font-size: 12px;
+}
+
+.overlay {
+  border: none !important;
+}
+
+.timeline {
```
jswk: This should go to map.component.scss, where the .timeline element is actually used. This applies to most of this file.
Please place the rules closer to the elements they are styling.

src/scss/bootstrap-view.scss
```diff
@@ -0,0 +1,94 @@
...
+  bottom: 20px;
+  left: 400px;
+  z-index: 100;
+}
+
+.product-view {
+  font-size: 13px;
+}
+.ol-zoom {
+  right: 9px !important;
```
jswk: !important isn't necessary here

sat4envi.iml
kosmidma: done

jswk: You didn't commit the removal of the file.

src/app/map/map.component.scss
```diff
@@ -1,22 +1,48 @@
...
   }
 }
+
+.timeline {
+  bottom: 20px;
+  left: 400px;
+  z-index: 100;
+}
+
+.ol-zoom {
```
jswk: This should go to the styles.scss file. When included here it doesn't have effect.

Issue #39, opened 2019-01-22
Overlay data model


jswk:
Add Overlays, WMSOverlays, PRGOverlays and PRGOverlayStatuses collections.

See https://docs.cyfronet.pl/pages/viewpage.action?pageId=46335820 for DB diagram.

PR #71, opened 2019-03-21
#39 Overlay data model

kmarszalek:
#39

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/bean/Overlay.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/bean/Overlay.java
@@ -1,0 +1,17 @@
+package pl.cyfronet.s4e.bean;
+
+import lombok.Data;
+
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.MappedSuperclass;
+
+@MappedSuperclass
+@Data
+public class Overlay {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String name;
+}
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/bean/PRGOverlay.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/bean/PRGOverlay.java
@@ -1,0 +1,12 @@
+package pl.cyfronet.s4e.bean;
+
+import lombok.Data;
+
+import javax.persistence.*;
+
+@Entity
+@Data
+public class PRGOverlay extends Overlay {
+    @ManyToOne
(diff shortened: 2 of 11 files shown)


kmarszalek:
its PR for discussion :)

s4e-backend/src/main/java/pl/cyfronet/s4e/bean/Overlay.java
```diff
@@ -0,0 +1,18 @@
...
+import javax.persistence.MappedSuperclass;
+
+@MappedSuperclass
+@Data
+public class Overlay {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String name;
+    private String projection;
```
jswk: After some thought `projection` isn't needed. It can be inferred based on the shapefiles by GeoServer itself when creating the layers and datastores.
Also, I wrote the DB diagram thinking about polar stereographic, now when we mostly default to Web Mercator this doesn't have to be stored independently in the DB.

s4e-backend/src/main/java/pl/cyfronet/s4e/bean/PRGOverlay.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Data;
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+
+@Entity
+@Data
+public class PRGOverlay extends Overlay {
```
jswk: There is one thing missing in the DB diagram that I realized is necessary recently: the SLD style of the layer.
It has to be customized from the GeoServer defaults and uploaded to the server independently.
An entity class should look like below. `name` for identification in GeoServer and `sldContent` for the xml style that has to be uploaded to the server.
```java
public class SldStyle {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;
    private String name;
    private String sldContent;
}
```
Then, `PRGOverlay` will need a field:
```java
    @ManyToOne
    private SldStyle sldStyle;
```

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/OverlayController.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @GetMapping("/overlays/prg/")
+    public List<PRGOverlayResponse> getPRGOverlays() {
+        return prgOverlayService.getPRGOverlays().stream()
+                .map(PRGOverlayResponse::of)
+                .collect(Collectors.toList());
+    }
+
+    @GetMapping("/overlays/wms/")
+    public List<WMSOverlay> getWMSOverlays() {
```
jswk: Return a list of WMSOverlayResponses

PR #81, opened 2019-04-04
[#34,#44] Register products in GeoServer instance. Integration tests. Spring-boot update

### LINKED ISSUE #34: Register granule with a GeoServer instance
When creating a product, register its appropriate Granule with a running GeoServer instance.

Blocked by: #33.

### LINKED ISSUE #44: Populate GeoServer instance with granules

jswk:
GeoServerOperations provides create, list and delete operations for
workspaces, coverage stores and S3 coverages.

GeoServerService so far is capable of resetting workspace and creating
coverage store and coverage (with layer) for a product. I added a field
Product.layerCreated to record if the layer has been created for it.

Also, Product has field s3Path added, which holds the path to granule.
I assume we will have a single s3 endpoint and bucket, so these are not
included in s3Path.

DBSeedDevelopment is updated to automatically populated GeoServer.
However, until PRG layers are also populated automatically they have to
be populated by scripts (a separate workspace is used for that).
The populating scripts have the s3 layer population removed.

I updated front-end to work with these changes.

I added some backend integration tests marked with tag `integration`.
These require minio and geoserver, so they are not run by `maven test`,
but can be executed with `maven verify`, which is described in README.

Also, spring-boot is updated to 2.1.4.RELEASE, this is a bugfix release
so it is done in a single run.

Fixes #34.
Fixes #44.


s4e-web/src/app/app.constants.ts
```diff
@@ -12,6 +13,7 @@ export function s4eConstantsFactory(): IConstants {
   return {
     projection: {toProjection: 'EPSG:3857', coordinates: [19, 52]},
     geoserverUrl: 'http://localhost:8080/geoserver/wms',
+    geoserverWorkspace: 'development',
```
michal-szostak: And what about cases where it will not be `development` ?

jswk: For now, it will fail. It should be populated using some configuration returned from backend. I thought to handle it in another issue

PR #95, opened 2019-05-06
87 ui axure

abacz:


s4e-web/src/app/components/login/login.component.ts
```diff
@@ -9,7 +9,7 @@ import {untilDestroyed} from 'ngx-take-until-destroy';
 import {disableEnableForm} from '../../utils/miscellaneous/miscellaneous';
 
 @Component({
-  selector: 's4e-login',
+  selector: '[s4e-login]',
```
jswk: Better leave the tag as the selector. The other components do it that way and I don't see a reason to do it differently here.
Also: [Angular style guide](https://angular.io/guide/styleguide#style-05-03)

s4e-web/src/app/views/map-view/product-picker/product-picker.component.ts
```diff
@@ -2,7 +2,7 @@ import {Component, EventEmitter, Input, OnInit, Output} from '@angular/core';
 import {ProductType} from '../state/product-type/product-type.model';
 
 @Component({
-  selector: 's4e-product-picker',
+  selector: '[s4e-product-picker]',
```
jswk: Better leave the tag as the selector. The other components do it that way and I don't see a reason to do it differently here.
Also: [Angular style guide](https://angular.io/guide/styleguide#style-05-03)

jswk:
Also, a note on commit messages. Leave an empty line after the first line. Then, it will be treated as heading and the rest as the details.

Issue #65, opened 2019-03-08
Populate state with incoming products


jswk:
The products should be discovered in the s3 storage and the db and GS populated.

jswk:
Closing this as #107 has been merged.

PR #107, opened 2019-05-27
[#65] Populate geoserver with Products from s3

kmarszalek:


s4e-backend/src/test/java/pl/cyfronet/s4e/service/ProductServiceTest.java
```diff
@@ -49,4 +57,51 @@ public void shouldReturnAllProducts() {
...
+
+        assertThat(productRepository.count(), is(equalTo(0L)));
+
+        productService.saveProduct(product);
+
+        assertThat(productRepository.count(), is(equalTo(1L)));
+    }
+
+    @Test
+    public void shouldBuildFromWebhook(){
```
jswk: Shouldn't S3Util be mocked here? It is unit tested in a separate test.

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/WebhooksController.java
```diff
@@ -0,0 +1,37 @@
...
+import pl.cyfronet.s4e.bean.Product;
+import pl.cyfronet.s4e.bean.Webhook;
+import pl.cyfronet.s4e.service.GeoServerService;
+import pl.cyfronet.s4e.service.ProductService;
+
+
+import static pl.cyfronet.s4e.Constants.API_PREFIX_V1;
+
+@RestController
+@RequestMapping(API_PREFIX_V1)
```
jswk: Wouldn't it make sense to expose this controller on a different prefix?
In this way it wouldn't be accesible from the outside, i.e. through front-end.

s4e-backend/src/main/java/pl/cyfronet/s4e/util/S3Util.java
```diff
@@ -0,0 +1,28 @@
+package pl.cyfronet.s4e.util;
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+@Service
+@RequiredArgsConstructor
```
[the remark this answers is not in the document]
kmarszalek: nope, it was removed but only after i made this PR :)

Issue #29, opened 2019-01-10
Registration


jswk:
Create `/register` endpoint.

Upon registration, an email must be sent with email verification, see #28. Until verified, account should be inactive.

Blocked by: #25.
Related: #26, #28, #27.

- [x] @jswk: backend
- [ ] @michal-szostak: frontend

jswk:
I moved the UI part to #115.

PR #110, opened 2019-06-06
[#29] Add user registration

jswk:
I added AppUser.enabled field.
It is now by default set to true, but once email confirmation is set up
it should be set to false, and set to true after the email is confirmed.

Also, AppUser.email is given a unique constraint to prevent duplicate
entries.
The JPA exception is mapped to AppUserCreationException in AppUserService.
ErrorHandler returns status 200 in such case to prevent email
enumeration attack.

I created AppUserController, which handles registration under `/register`
endpoint.

Register- and LoginRequest get validation constraints.
Password for now only has length constraint.

The validation errors are handled in ErrorHandler and mapped to a useful
form using ErrorHandlerHelper.
The errors are translated to Polish using ResourceBundleMessageSource
and a file with translations.

AppRoleSetConverted has added tests, as it turned out it failed if
an AppUser had no roles.

Fixes: #29.

@michal-szostak: have a look at how the error response is constructed in ErrorHandlerHelper
@kmarszalek: have a look at the rest


jswk:
@michal-szostak I merged this PR to work on further functionality on the back-end already. On the other hand I opened a separate #115 for the UI.

Issue #28, opened 2019-01-10
Email verification


jswk:
Set up email verification. An email with a token should be issued. Keep the token and new email in a separate table.
API should expose an endpoint `/confirm-email?token=<token>`. When the right token is entered AppUser.email should be updated, and the account activated (if not already).
The token will have an expiration time.

If expired, the user can trigger sending another verification mail. It is done when the user attempts to access the original verification url.

See https://www.baeldung.com/registration-verify-user-by-email.

**UPDATE** (by @michal-szostak)

After some discussions we decided to change a scope of the issue a little bit. The **additional** requirements are as follows:

* LoginController should respond with different error codes depending on whether user account has been activated or not:
   * `401` for wrong credentials
   * `403` for correct credentials but disabled (not activated) user account
* Additional endpoint - `POST` input `activation token` action `resend verification email for the user for which provided activation tokan was provided`

**UPDATE** (by @jswk)

The `/confirm-email` endpoint will return:
* `200` if email has been successfully confirmed
* `401` if the token has expired
* `404` if the token was not found

Related: #27.

PR #117, opened 2019-07-02
[#28] Email verification

jswk:
Add token based email verification. An AppUser is created with field
enabled=false, and it can be activated using the token sent by mail.

The logic of sending emails is decoupled from the controller to async
event listeners.

Create LoggingMailSender, which outputs email contents to log.

For testing I introduced GreenMail, which is used to verify appropriate
emails are sent.

The email contents are put in messages_pl_PL.properties.

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/controller/AppUserController.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/controller/AppUserController.java
@@ -1,19 +1,29 @@
 package pl.cyfronet.s4e.controller;
 
+import io.swagger.annotations.*;
 import lombok.RequiredArgsConstructor;
+import lombok.val;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.i18n.LocaleContextHolder;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.crypto.password.PasswordEncoder;
-import org.springframework.web.bind.annotation.PostMapping;
-import org.springframework.web.bind.annotation.RequestBody;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.*;
 import pl.cyfronet.s4e.bean.AppRole;
 import pl.cyfronet.s4e.bean.AppUser;
 import pl.cyfronet.s4e.controller.request.RegisterRequest;
+import pl.cyfronet.s4e.event.OnEmailConfirmedEvent;
+import pl.cyfronet.s4e.event.OnRegistrationCompleteEvent;
+import pl.cyfronet.s4e.event.OnResendRegistrationTokenEvent;
 import pl.cyfronet.s4e.ex.AppUserCreationException;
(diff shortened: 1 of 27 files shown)


michal-szostak:
I updated issue scope with new requirements: #28

jswk:
Another update in #28.

jswk:
I updated the code to match the updated requirements. @michal-szostak please verify if it works as expected.

s4e-backend/src/main/java/pl/cyfronet/s4e/bean/RegistrationToken.java
```diff
@@ -0,0 +1,24 @@
...
+
+import javax.persistence.*;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import java.time.LocalDateTime;
+
+@Entity
+@Data
+@Builder
+public class RegistrationToken {
```
kmarszalek: not emailverification?

jswk: I was considering EmailVerificationToken and it was too long for me. But actually just EmailVerification sounds better. The token is a component of such verification then. I'll refactor it.

s4e-backend/src/main/java/pl/cyfronet/s4e/service/AppUserService.java
```diff
@@ -24,4 +26,8 @@ public AppUser save(AppUser appUser) throws AppUserCreationException {
             throw new AppUserCreationException(e);
         }
     }
+
+    public Optional<AppUser> findByEmail(String email) {
```
kmarszalek: should we return from services one type of return, optional or e.g. AppUser/boolean ?

s4e-backend/src/main/java/pl/cyfronet/s4e/service/VerificationTokenService.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.temporal.TemporalAmount;
+import java.util.Optional;
+import java.util.UUID;
+
+@Service
+@RequiredArgsConstructor
+public class VerificationTokenService {
```
kmarszalek: same as other service :)

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/AppUserController.java
```diff
@@ -22,20 +32,87 @@
...
+                    @ExampleProperty(mediaType = "application/json", value = "{\"email\":[\"musi być adresem e-mail\"]}")
+            }))
+    })
     @PostMapping("/register")
-    public ResponseEntity<?> register(@RequestBody @Valid RegisterRequest registerRequest) throws AppUserCreationException {
-        appUserService.save(AppUser.builder()
+    public ResponseEntity<?> register(
+            @RequestBody @Valid RegisterRequest registerRequest
+    ) throws AppUserCreationException {
+        AppUser appUser = appUserService.save(AppUser.builder()
```
kmarszalek: check if exists? repo.save do save or update...

Issue #120, opened 2019-07-04
reCAPTCHA backend


jswk:
Add reCAPTCHA v2 support to /register endpoint.

Use [recaptcha-spring-boot-starter](https://github.com/mkopylec/recaptcha-spring-boot-starter) to verify the request.
Add info about passing 'g-recaptcha-response' in the request. Use a request parameter for that.
Header could also be used, but a parameter should be ok too and the parameter can have the original name, i.e. 'g-recaptcha-response' and the boot starter handles this case out of the box.

The site-key should be passed to the front-end by configuration endpoint #88, #89.

The verification error should end with getting an error description under `__general__`.

PR #123, opened 2019-07-05
[#120] reCAPTCHA backend

jswk:
Add dependency to recaptcha-spring-boot-starter, which handles
sending request to the Google reCAPTCHA service.

The RecaptchaValidator is called in the AppUserController::register,
validating the received param `g-recaptcha-response`.

In case of validator error appropriate error codes are returned in
field `recaptcha`.

For tests the validator test mode is enabled, and in development it
uses the always-returning-true key-set which is recommended for tests in
reCAPTCHA FAQ.

Add field `recaptchaSiteKey` to the `/api/v1/config` response.

Fixes: #120.

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/db/SeedDevelopment.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/db/SeedDevelopment.java
@@ -31,6 +31,7 @@ public class SeedDevelopment implements ApplicationRunner {
     private final SldStyleRepository sldStyleRepository;
     private final PRGOverlayRepository prgOverlayRepository;
     private final AppUserRepository appUserRepository;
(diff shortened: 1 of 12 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/db/SeedDevelopment.java
```diff
@@ -39,6 +40,7 @@
     @Async
     @Override
     public void run(ApplicationArguments args) {
+        emailVerificationRepository.deleteAll();
```
kmarszalek: you. are. lazy.

jswk: thank you!

Issue #32, opened 2019-01-10
Reset password


jswk:
Allow user to reset her password.
Upon providing an email or a login user is sent an email with a password reset token (subject to expiry).
On entering the link the user can set a new password and authenticate with it.

See: https://www.baeldung.com/spring-security-registration-i-forgot-my-password.

Related: #31, #26, #51.

PR #124, opened 2019-07-12
#32 Password reset with token

kmarszalek:


s4e-backend/src/main/resources/db/migration/V4__add_resetl_password_table.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE password_reset (
+    id BIGSERIAL PRIMARY KEY,
+    app_user_id BIGSERIAL UNIQUE REFERENCES app_user NOT NULL,
+    token VARCHAR UNIQUE NOT NULL,
+    expiry_timestamp TIMESTAMP NOT NULL,
+    enabled BOOLEAN DEFAULT FALSE
```
jswk: In the bean the field is called `used`

s4e-backend/src/main/resources/db/migration/V4__add_resetl_password_table.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE password_reset (
```
jswk: A typo in the filename. `l` after `reset`. Also, the filename should read `password_reset` and not `reset_password`.

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/PasswordController.java
```diff
@@ -0,0 +1,91 @@
...
+    private final AppUserService appUserService;
+    private final PasswordService passwordService;
+    private final ApplicationEventPublisher eventPublisher;
+
+    @ApiOperation("Create and send a password reset token based on email")
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "Email with a password reset token was sent provided user with it existed")
+    })
+    @PostMapping(value = "/token-create")
+    public ResponseEntity<?> resetPassword(@RequestParam @Email @NotEmpty @Valid  String email) {
```
jswk: extra space after annotations

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/PasswordController.java
```diff
@@ -0,0 +1,91 @@
...
+    @PostMapping(value = "/password-reset")
+    @ResponseBody
+    public ResponseEntity<?> resetPassword(@RequestBody @Valid PasswordResetRequest passwordReset, @RequestParam("token") String token) throws NotFoundException, BadRequestException {
+        val optionalResetPasswordToken = passwordService.findByToken(token)
+                .orElseThrow(() -> new NotFoundException("Provided token '" + token + "' not found"));
+        if (optionalResetPasswordToken.getAppUser().getPassword().equals(passwordReset.getOldPassword())) {
+            AppUser user = optionalResetPasswordToken.getAppUser();
+            user.setPassword(passwordReset.getNewPassword());
+            appUserService.update(user);
+            optionalResetPasswordToken.setUsed(true);
```
jswk: Do you check if it was used when actually resetting the password? I only see the check in `/token-validate`.
Also, shouldn't the token just be deleted after it is used? Currently, if I understand the code correctly, PasswordService won't be able to generate a second token for an AppUser. Maybe user could generate another token even if he has one already?

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/PasswordController.java
```diff
@@ -0,0 +1,116 @@
...
+            @ApiResponse(code = 404, message = "The user was not found")
+    })
+    @PostMapping(value = "/password-change")
+    @ResponseBody
+    @PreAuthorize("!isAnonymous()")
+    public ResponseEntity<?> changePassword(@RequestBody @Valid PasswordResetRequest passwordReset) throws NotFoundException, BadRequestException {
+        val optionalAppUser = appUserService.findByEmail(
+                ((AppUserDetails) SecurityContextHolder.getContext().getAuthentication().getDetails()).getAppUser().getEmail())
+                .orElseThrow(() -> new NotFoundException());
+        if (optionalAppUser.getPassword().equals(passwordReset.getOldPassword())) {
```
jswk: Actually, this check will always fail. The password in AppUser is hashed. Use PasswordEncoder to encode the password from PasswordResetRequest before checking for equality.

PR #129, opened 2019-07-30
Move to OAuth2.0

jswk:
Remove the pre-existing login and JWT related classes.

Add authorization and resource server configurations.
For now no refresh token revocation is implemented, as the JwtTokenStore
doesn't store tokens. Moreover, the revocation endpoint is not a part of
the spring-security-oauth and must be implemented on its own.


jswk:
Official docs:
https://projects.spring.io/spring-security-oauth/docs/oauth2.html

A guide to which grant type to use:
https://alexbilbie.com/guide-to-oauth-2-grants/

A tutorial on how to secure API. Consider looking at previous parts as well:
https://sivatechlab.com/secure-rest-api-using-spring-security-oauth2-part-3/

On using JWT:
https://www.baeldung.com/spring-security-oauth-jwt

Issue #135, opened 2019-08-26
Provision GeoServer with more data


jswk:
Use the existing data in loop to generate data in db and GS for about 2 months worth of geostationary products.

jswk:
This will be done in a different fashion in #200.

PR #137, opened 2019-08-27
[#135] Generate a month of products

jswk:
The data from a single day is repeated, generating data for 30 days in SeedDevelopment.
This requires around 2k layers for 3 product types, which takes several minutes. The progress is indicated in the logs.


jswk:
Closing it, as it doesn't necessarily has to be a PR.
It's enough that @michal-szostak can use this branch for timeline development.

Issue #141, opened 2019-09-03
Account Activation


michal-szostak:
Create view for account activation

PR #142, opened 2019-09-03
[#141] Account Activation

tsatlawa:


tsatlawa:
Squashes and repushed, please rev again

Issue #192, opened 2019-11-05
DateConverter tests fail


michal-szostak:
### Error log

```
[ERROR]
[ERROR]   ● DateConverter › should deserialize
[ERROR]
[ERROR]     expect(received).toEqual(expected)
[ERROR] 
[ERROR]     Difference:
[ERROR] 
[ERROR]     - Expected
[ERROR]     - 2019-03-21T22:04:19.000Z
[ERROR]     + Received
[ERROR]     + 2019-03-21T23:04:19.000Z
[ERROR] 
[ERROR] 
[ERROR]       33 |   it('should deserialize', () => {
[ERROR]       34 |     expect(service.deserialize('2019-03-21T23:04:19'))
[ERROR]     > 35 |       .toEqual(new Date(Date.UTC(2019, 2, 21, 22, 4, 19, 0)));
[ERROR]          |        ^
[ERROR]       36 |   });
[ERROR]       37 | });
[ERROR]       38 | 
```

### Potential solution

The most probable source of the issue is using timezone marked dates instead of UTC ones

PR #193, opened 2019-11-05
[#192] DateConverter tests fail

michal-szostak:
## What is in this PR?

* move to moment js, which handles UTC dates much better than date-fms

## How to test whether it works?

If the travis / github pipeline tests do not fail for frontend then it means that the fix was successful

Closes: #192


michal-szostak:
Ok, got it, I havent't checked all tests in all test suites, will fix it

Issue #168, opened 2019-10-08
Institution and group management back-end


jswk:
Beans for institutions, CRUD. Permissions.

kmarszalek:
Permissions in the scope of this tickets will be write down, roles #169 and its usage via annotation on rest controller level will implement this feature.

bwilk:
### Endpoints
#### GET Profile 
Needed to show the profile page 
- user name, surname, email  
- user roles (eg. { role: ADMIN }, { role: MANAGER, group: 11 },  { role: ADMIN, institution: 2 }, { role: USER, institution: 2 } )

#### CREATE user with roles 
Parameters:
 - user name, surname, email
 - grup_ids
Endpoint should authorize the requesting user with roles in groups (by group_ids)

Enhance payload of endpoints:
- POST, PUT group

kmarszalek:
last point:
CREATE user with roles
Parameters:

user name, surname, email
grup_ids -> groupsSlugs :)

kmarszalek:
get profile, add user and enhance post and put for groups done.
put for user in institution, batch operation o groups for user in scope of institution.

PR #206, opened 2019-11-08
[#168] Add endpoint for User profile

kmarszalek:


s4e-backend/src/main/java/pl/cyfronet/s4e/controller/AppUserController.java
```diff
@@ -142,11 +145,26 @@
...
+    @ApiOperation("Get user profile")
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "User was added"),
+            @ApiResponse(code = 403, message = "User not authorized"),
+            @ApiResponse(code = 404, message = "User wasn't found")
+    })
+    @GetMapping("/users/me")
+    @PreAuthorize("isAuthenticated()")
+    public AppUserResponse getMe() throws NotFoundException {
+        UserDetails userDetails = (UserDetails) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
```
jswk: Why not cast to AppUserDetails and extract AppUser?

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/AppUserController.java
```diff
@@ -142,11 +145,26 @@
...
+        val optionalUser = appUserService.findByEmail(userDetails.getUsername())
```
jswk: Also, this is not optional

kmarszalek: how so? AppUserService::Optional<AppUser> findByEmail(String email)

Issue #200, opened 2019-11-07
Prepare data set for SOK tutorial


jswk:
Create another data pack for use during the upcoming SOK tutorial.
It should consist of products already on Prometheus, with around a month of data for each. Time resolution 1h, can be even 2h.

PR #207, opened 2019-11-08
[#200] Products seed for s4e-demo data set

jswk:
Rework the SeedProducts to enable multiple data set support. For now
only two are available, so I don't introduce any extra abstraction over
the choice mechanism. However, I introduce a parameterized product
generation.

Improve GeoServerOperations timeout handling, which is now set to 60s
and can be customized in case of problems.

The GeoServerOperations::layerExists didn't work as expected, I now
catch the NotFound exception and translate it to false output.

Remove unused GeoServerSychronizer methods.

Closes: #200.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,10 +80,83 @@ Several profiles can be specified in a comma separated list, e.g. `production,ru
 `run-seed-users`: Runs `SeedUsers`.
 
 `run-seed-products`: Runs `SeedProducts`.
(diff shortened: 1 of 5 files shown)


michal-szostak:
I checked it out and it works, but I would recommend adding separate entry in `README.md` explaining step by step how to turn this feature on. Otherwise I approve

PR #213, opened 2019-11-12
[#100] Places search front-end

### LINKED ISSUE #100: Places search front-end
Use the API created in #46 (PR: #99) to fetch search results to the front-end.
The `/places` endpoint returns paged results with JSON resembling the one presented [here](https://wimdeblauwe.wordpress.com/2018/06/10/pageimpl-json-serialization-with-spring-boot-2/) (for Spring Boot 2.0.2).

After selecting a result the map should center at it.

tsatlawa:
* Query backend for places
* Display results in SearchResultComponent
* Mark selected place with pin

Closes #100

### DIFF
--- a/s4e-web/src/app/views/map-view/map/map.component.ts
+++ b/s4e-web/src/app/views/map-view/map/map.component.ts
@@ -1,14 +1,19 @@
 import {Component, Input, OnDestroy, OnInit} from '@angular/core';
 import Map from 'ol/Map';
 import View from 'ol/View';
-import {Image, Layer, Tile} from 'ol/layer';
-import {ImageWMS, OSM} from 'ol/source';
+import Feature from 'ol/Feature';
+import {Image, Layer, Tile, Vector as VectorLayer} from 'ol/layer';
+import {ImageWMS, OSM, Vector} from 'ol/source';
+import {Icon, Style} from 'ol/style';
 import {UIOverlay} from '../state/overlay/overlay.model';
 import proj4 from 'proj4';
 import {Product} from '../state/product/product.model';
-import {ReplaySubject} from 'rxjs';
+import {Observable, ReplaySubject, Subscription} from 'rxjs';
 import {untilDestroyed} from 'ngx-take-until-destroy';
 import {S4eConfig} from '../../../utils/initializer/config.service';
+import {SearchResult} from '../state/search-results/search-result.model';
+import {Point} from 'ol/geom';
+
 
 @Component({
   selector: 's4e-map',
@@ -17,10 +22,12 @@ import {S4eConfig} from '../../../utils/initializer/config.service';
 })
 export class MapComponent implements OnInit, OnDestroy {
   @Input() public overlays: UIOverlay[] = [];
+  function;
+  selectedLocationSub: Subscription = null;
   private baseLayer: Layer;
   private map: Map;
-
   private activeProduct$ = new ReplaySubject<Product | null>(1);
+  private markerSource: Vector = new Vector();
 
   constructor(private CONFIG: S4eConfig) {
   }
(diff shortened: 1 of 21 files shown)


s4e-web/src/app/views/map-view/map/map.component.ts
```diff
@@ -52,6 +88,21 @@ export class MapComponent implements OnInit, OnDestroy {
...
+        src: 'https://cdn2.iconfinder.com/data/icons/ui-26/128/map-pin-location-256.png'
+      }))
+    });
+    this.map.getLayers().push(new VectorLayer({
+      source: this.markerSource,
+      style: markerStyle,
+    }));
   }
 
   ngOnDestroy(): void {
```
michal-szostak: You should call `this.selectedLocationSub.unsubscribe();` in `ngOnDestroy()` as well as nullify `this.selectedLocationSub`

s4e-web/src/app/views/map-view/search-results/search-results.component.html
```diff
@@ -0,0 +1,22 @@
+<div class="searchResult" >
+
+  <div [hidden]="!loading">
+    <fa-icon [icon]="['fas', 'circle-notch']" [spin]="true"></fa-icon>
+    Ładowanie wyników
```
michal-szostak: string literals should be wrappen in some tags with `i18n` attribute

s4e-web/src/app/views/map-view/search-results/search-results.component.html
```diff
@@ -0,0 +1,22 @@
...
+  </div>
+
+  <div [hidden]="loading">
+    <div *ngIf="searchResults.length == 0">
+      Brak wyników
```
michal-szostak: string literals should be wrappen in some tags with `i18n` attribute

s4e-web/src/app/views/map-view/view-manager/view-manager.component.ts
```diff
@@ -1,23 +1,56 @@
...
+
+  ngOnInit(): void {
+    if (environment.hmr) {
+      this.searchFc.setValue(this.searchResultQuery.getValue().queryString);
+    }
+
+    this.searchFc.valueChanges.pipe(
+      debounceTime(300),
+      distinctUntilChanged()
+    ).subscribe((text: string) => this.searchForPlaces.emit(text));
```
michal-szostak: Subscriptions must be unsubscribed in `ngOnDestroy`, it can be easily done (without storing subscription reference using a helper `untilDestroy`:

```
    this.searchFc.valueChanges.pipe(
        debounceTime(300),
        distinctUntilChanged(),
        untilDestroyed(this),
      ).subscribe((text: string) => this.searchForPlaces.emit(text));
```

check out usage in other components, that should provide you with necessary information

s4e-web/src/app/views/map-view/state/search-results/search-results.service.ts
```diff
@@ -0,0 +1,35 @@
...
+export class SearchResultsService {
+
+  constructor(private searchResultsStore: SearchResultsStore,
+              private http: HttpClient,
+              private CONFIG: S4eConfig) {
+  }
+
+  get(placePrefix: string) {
+    this.searchResultsStore.update({isOpen: placePrefix.length > 0});
+    this.searchResultsStore.update({queryString: placePrefix});
```
michal-szostak: This can be done using one call to `.update`:

```
 this.searchResultsStore.update({isOpen: placePrefix.length > 0, queryString: placePrefix});
```

s4e-web/src/app/views/map-view/state/search-results/search-results.service.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      req.flush({
+        content: [sampleResult]
+      });
+      tick(1000);
+
+      expect(searchResultsQuery.getValue().isOpen).toEqual(true);
+      expect(searchResultsQuery.getValue().queryString).toEqual('k');
+      const ids = searchResultsQuery.getValue().ids;
+      expect(ids.length).toEqual(1);
+      sampleResult['id'] = ids[0]; // ??? ASK MICH
```
michal-szostak: We talked about it, but I'll repost it here for completeness sake:
I would advice to mock `guid()` via spyOn. I'm not 100% sure it'll work without hitch, as mocking methods from other modules is notoriously problematic. If there is a problem with mocking it you could try extracting call to `guid()` in service to separate service's private method, and mock that one.
It kind of breaks the idea that tests should not affect code structure, but I think it's a reasonable price to pay in this case.

s4e-web/src/app/views/map-view/state/search-results/search-results.query.ts
```diff
@@ -0,0 +1,18 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class SearchResultsQuery extends QueryEntity<SearchResultsState, SearchResult> {
+
+  constructor(protected store: SearchResultsStore) {
+    super(store);
+  }
+
+  selectIsOpen() {
```
michal-szostak: Providing return type is a good practice. I know, that it can be inferred from the function call, but it's a good practice nevertheless

Issue #220, opened 2019-11-13
Lazy loading


kmarszalek:
rethink lazyloading, maybe projection or jpa query?

Issue #169, opened 2019-10-08
Roles system


jswk:
Roles for admins, institutions, group managers.
CRUD with security wired up in controllers.
Write down the planned model in confl.

PR #227, opened 2019-11-15
[#169] Secure endpoints for Institutions, Groups and Profile

kmarszalek:
- add @PreAuthorize("isAuthenticated()") annotations and tests adjustment

changes with configuration of  groupControllerTest required for multiple mockMvc.perform methods inside one @Test method. should be refactored in scope of #220.


s4e-backend/src/test/java/pl/cyfronet/s4e/controller/GroupControllerTest.java
```diff
@@ -54,149 +58,149 @@
 
     private String slugInstitution = "";
 
+    public static final String profileEmail = "get@profile.com";
```
jswk: UPPERCASE property name?
Also, why not use a constant in other tests?

PR #261, opened 2019-11-27
[#257] Save PNG of the map canvas

### LINKED ISSUE #257: Generating PNG image from canvas
This issue contains two parts:

1. [ ] Research way to save canvas state to png (getting pixel matrix from canvas, writing png file through JS library, etc
2. [ ] Connect that functionality to the button in the front end

michal-szostak:
## What is in this PR?

* take canvas data as png and save it
* add npm tasks to run production test server

## How to test?

Clicking placeholder button in the bottom right corner should download file

image

## Potential problems

This solution will work only in newer browsers. 
@abacz - question - what should be name of the saved file? Now I've set it to be `SNAPSHOT.<DATE>.png` will it suffice?

Closes #257


tsatlawa:
For me it is working after redux clear and when I select products. However, selecting overlay causes following error:
On chrome (Wersja 78.0.3904.108 (Oficjalna wersja) (64-bitowa))
`ERROR DOMException: Failed to execute 'toDataURL' on 'HTMLCanvasElement': Tainted canvases may not be exported.`
On firefox (70.0.1 (64 bity))
`ERROR DOMException: "The operation is insecure."`

michal-szostak:
@tsatlawa - I fixed that error, it turned out to be an error with CORS (drawing image loaded with `no-cors` policy tainted canvas, and for security reasons loading its state was being prevented by the browser. I changed proxies, cors, and a couple parameters and it should be alright now, but it's necessary to recompile backend

PR #265, opened 2019-12-03
[#254] Drop Down for ZK

### LINKED ISSUE #254: Drop Down (show / hide) for ZK options
It's still undecided whether it will be drop down on the navbar (which does not exist at the moment) or element sliding from the right (_drawer_), the component itself should only show / hide itself based on the clicking of the button

michal-szostak:
## What is in this PR?

* Add dropdown for ZK to MapViewComponent
* Add tests to MapViewComponent
* Change z-index of modals, to make them always on top

## How to test / how does it look like?

Screenshot 2019-12-03 at 15 41 33

Screenshot 2019-12-03 at 15 41 47

Closes #254

### DIFF
--- a/s4e-web/src/app/views/map-view/map-view.component.html
+++ b/s4e-web/src/app/views/map-view/map-view.component.html
@@ -12,8 +12,7 @@
                   [searchResults]="placeSearchResults$ | async"
                   [searchResultsLoading]="placeSearchLoading$ | async"
                   [searchResultsOpen]="placeSearchResultsOpen$ | async"
-                  (searchResultClicked)="navigateToPlace($event)"
->
+                  (searchResultClicked)="navigateToPlace($event)">
 </s4e-view-manager>
 <s4e-legend *ngIf="legend$ | async"
             [isOpen]="(legendState$ | async).isOpen"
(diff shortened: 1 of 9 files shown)


s4e-web/src/app/views/map-view/state/map/map.service.ts
```diff
@@ -14,4 +14,8 @@ export class MapService {
               private CONFIG: S4eConfig
   ) {
   }
+
+  toggleZKOptions(open: boolean = true) {
```
jswk: Shouldn't it be called `showZKOptions` or `setZKOptionsOpened`? Toggle may be misleading

s4e-web/src/app/views/map-view/map-view.component.html
```diff
@@ -34,9 +33,25 @@
...
          [selectedLocation]="selectedLocation$"
          [overlays]="overlays$ | async"
-         [activeScene]="activeScene$ | async"
->
-
+         [activeScene]="activeScene$ | async">
 </s4e-map>
+
+<ng-container *ngIf="userLoggedIn$ | async">
+  <button (click)="toggleZKOptions()" id="zk-options-button" style="position:absolute; right: 60px; top: 15px;" class="button button--default">
```
jswk: The styles may be better to keep in `map-view.component.scss`?

PR #268, opened 2019-12-03
[#231] Rename Product and ProductType

jswk:
## What is in this PR?

* Rename Product to Scene
* Rename ProductType to Product
* Fix foreign keys default values

The changes are done in separate commits to contain the rename better.

## How to test?

Run the application in development.
The front-end should work as before, ie. load products, scenes, show them etc.
The front- and back-end tests should pass, I tried to rename variables and names where possible, if I missed something give me heads up.
@michal-szostak please have a look if I didn't mess something up in the front-end

The db is migrated to the new names, the names of the tables and columns reflect the changes - check if they are correct.
Moreover, notice that a number of sequences is reduced - foreign keys were incorrectly defined as BIGSERIAL, which resulted in unused sequences.

Closes: #231.


s4e-backend/src/main/java/pl/cyfronet/s4e/controller/SceneController.java
```diff
@@ -0,0 +1,57 @@
...
+        return sceneService.getScenes(productId).stream()
+                .map(SceneResponse::of)
+                .collect(Collectors.toList());
+    }
+
+    @ApiOperation(value = "Return days on which Product is available")
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "Successfully retrieved list")
+    })
+    @GetMapping("/scenes/productId/{productId}/available")
```
kmarszalek: and here.
not the scope of this PR but, shouldnt this endpoint look like: products/{productId}/scenes/available :)

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/SceneController.java
```diff
@@ -0,0 +1,57 @@
...
+@RequestMapping(API_PREFIX_V1)
+@RequiredArgsConstructor
+public class SceneController {
+    private final SceneService sceneService;
+
+    @ApiOperation(value = "View a list of scenes")
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "Successfully retrieved list")
+    })
+    @GetMapping("/scenes/productId/{productId}")
```
kmarszalek: not the scope of this PR but, shouldnt this endpoint look like: products/{productId}/scenes :)

Issue #267, opened 2019-12-03
Migrate API docs to springdoc-openapi


jswk:
Springfox seems to be effectively dead right now, the correct way would be to migrate to its alternative: springdoc-openapi.

PR #280, opened 2019-12-10
[#267] Move API docs to springdoc-openapi

jswk:
Closes: #267.


s4e-backend/src/main/java/pl/cyfronet/s4e/OpenApiConfig.java
```diff
@@ -0,0 +1,14 @@
+package pl.cyfronet.s4e;
+
+import io.swagger.v3.oas.annotations.OpenAPIDefinition;
+import io.swagger.v3.oas.annotations.info.Info;
+
+@OpenAPIDefinition(
```
kmarszalek: maybe move to Application? :)

PR #293, opened 2019-12-23
249 share link

kmarszalek:
it waits for UserRole PR to be merged.

input:
 - link to share without hostname
 - list of emails

only ZKMembers can access this endpoint
backend have properties with front base host

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/controller/ShareLinkController.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/controller/ShareLinkController.java
@@ -1,0 +1,58 @@
+package pl.cyfronet.s4e.controller;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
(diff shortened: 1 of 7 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/controller/ShareLinkController.java
```diff
@@ -0,0 +1,58 @@
...
+    private final ApplicationEventPublisher eventPublisher;
+    private final AppUserService appUserService;
+
+    @Operation(
+            summary = "Share link",
+            description =
+                    "Send email with shared linked for user of ZK to access"
+    )
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Operation successful"),
```
jswk: Also has no content when 200. Add `content = @Content`?

s4e-backend/src/main/java/pl/cyfronet/s4e/listener/GroupListener.java
```diff
@@ -20,6 +22,9 @@
     private final TemplateEngine templateEngine;
     private final MailService mailService;
 
+    @Value("${front.web.base.url}")
```
jswk: Shouldn't this be `mail.urlDomain`? It's the same, and it's used in the same context of mail-sending.

Issue #313, opened 2020-01-10
Raport modal [HTML/CSS]


abacz:
Modal raport -> https://invis.io/S9RJNE8WXUB#/400043352_ZK_-_Modal_Raport

PR #318, opened 2020-01-13
[#313] Report modal html/css

abacz:


michal-szostak:
I think it's ok. There are two cases regarding scaling of textarea by the user, and scrolling modals (see attached images). But those issues can be resolved in another PR.

image

image-2

PR #322, opened 2020-01-15
[#243] refactor controllers

kmarszalek:


s4e-backend/src/main/java/pl/cyfronet/s4e/controller/response/AppUserResponse.java
```diff
@@ -1,30 +1,14 @@
...
-    private String name;
-    private String surname;
-    private boolean admin;
-    private boolean memberZK;
-    private Set<UserRoleResponse> roles;
+public interface AppUserResponse {
+    String getEmail();
+
+    String getName();
+    boolean getAdmin();
```
jswk: Whitespace

s4e-backend/src/main/java/pl/cyfronet/s4e/listener/GroupListener.java
```diff
@@ -27,17 +28,18 @@
...
     private final TemplateEngine templateEngine;
     private final MailService mailService;
+    private final GroupService groupService;
 
     @Value("${mail.urlDomain}")
     private String urlDomain;
 
     @Async
     @EventListener
+    @Transactional(readOnly = true)
```
jswk: Why these `@Transactional`?

s4e-backend/src/main/java/pl/cyfronet/s4e/service/AppUserService.java
```diff
@@ -57,6 +79,26 @@ public AppUser createFromRequest(CreateUserWithGroupsRequest request, String ins
         return appUser;
     }
 
+    public EmailVerification confirmEmail(String token) throws NotFoundException, RegistrationTokenExpiredException, AppUserCreationException {
```
jswk: You should declare `@Transactional`, the same as in ::save, as you call it from within the class (and so it won't have a transaction open). Moreover, it might make sense not to call ::save here, but rely on the hibernate for saving the entity as you are only updating it (the `@Transactional` is then essential here).

s4e-backend/src/main/java/pl/cyfronet/s4e/service/EmailVerificationService.java
```diff
@@ -18,16 +20,19 @@
...
+    public Optional<EmailVerification> findByAppUserEmail(String email) {
+        return emailVerificationRepository.findByAppUserEmail(email);
     }
 
     public Optional<EmailVerification> findByToken(String token) {
         return emailVerificationRepository.findByToken(token);
     }
 
-    public EmailVerification create(AppUser appUser) {
+    public EmailVerification create(String email) throws NotFoundException {
```
jswk: This should probably have `@Transactional`?

s4e-backend/src/main/java/pl/cyfronet/s4e/service/PasswordService.java
```diff
@@ -18,13 +27,52 @@
 @Slf4j
 public class PasswordService {
     private static final TemporalAmount EXPIRE_IN = Duration.ofDays(1);
+    private final AppUserService appUserService;
+    private final PasswordEncoder passwordEncoder;
     private final PasswordResetRepository passwordResetRepository;
 
-    public PasswordReset createPasswordResetTokenForUser(AppUser user) {
-        PasswordReset token = PasswordReset.builder().token(UUID.randomUUID().toString()).appUser(user).expiryTimestamp(LocalDateTime.now().plus(EXPIRE_IN)).build();
+    public PasswordReset createPasswordResetTokenForUser(String email) throws NotFoundException {
```
jswk: `@Transactional`?

s4e-backend/src/main/java/pl/cyfronet/s4e/service/UserRoleService.java
```diff
@@ -21,15 +23,19 @@
     private final AppUserRepository appUserRepository;
     private final GroupRepository groupRepository;
 
+    @Transactional
+    public void addRole(CreateUserRoleRequest request) throws NotFoundException {
+        addRole(request.getRole(), request.getEmail(), request.getInstitutionSlug(), request.getGroupSlug());
+    }
+
     @Transactional
```
jswk: Don't you need to specify the `NotFoundException` here? (`rollbackFor`, or sth like this)

s4e-backend/src/main/java/pl/cyfronet/s4e/listener/GroupListener.java
kmarszalek: because institution in group is lazy and i use here normal method from repo, without query

s4e-backend/src/main/java/pl/cyfronet/s4e/service/UserRoleService.java
kmarszalek: why should i?

PR #323, opened 2020-01-15
[#321] Add extra share link fields

jswk:
Add caption, description and thumbnail fields to `/share-link` endpoint.
Also, rename link to path.

Rework the way the request parameters are handled and passed to the
event listener.

Exclude an unnecessary junit-vintage dependency from
spring-boot-starter-test.

Closes: #321.

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/controller/ShareLinkController.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/controller/ShareLinkController.java
@@ -6,23 +6,20 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import lombok.RequiredArgsConstructor;
+import lombok.val;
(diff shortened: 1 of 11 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/controller/ShareLinkController.java
```diff
@@ -34,25 +32,27 @@
...
-        eventPublisher.publishEvent(new OnShareLinkEvent(appUser, request.getLink(), request.getEmails(), LocaleContextHolder.getLocale()));
-        return ResponseEntity.ok().build();
+        val shareRequest = OnShareLinkEvent.Request.builder()
+                .caption(request.getCaption())
+                .description(request.getDescription())
+                .thumbnail(Base64.getDecoder().decode(request.getThumbnail()))
+                .path(request.getPath())
+                .emails(request.getEmails())
+                .build();
+        eventPublisher.publishEvent(new OnShareLinkEvent(appUserDetails, shareRequest, LocaleContextHolder.getLocale()));
```
kmarszalek: so only email :)

Issue #315, opened 2020-01-10
Improve password reset handling


jswk:
Allow user to send a reset token several times. Now, attempt to send more than one token effects in unique constraint violation. Remove the pre-existing token for a user.

Also, verify that the token has not expired when using it.

PR #346, opened 2020-01-23
[#315] Reset password handlig fix

kmarszalek:
- add deleting preexisting token while requesting a new one
- add token expiry verification
- add tests
- removed unnecessary content type in tests on get endpoints

### DIFF
--- a/s4e-backend/src/test/java/pl/cyfronet/s4e/service/PasswordServiceTest.java
+++ b/s4e-backend/src/test/java/pl/cyfronet/s4e/service/PasswordServiceTest.java
@@ -1,0 +1,239 @@
+package pl.cyfronet.s4e.service;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.crypto.password.PasswordEncoder;
+import pl.cyfronet.s4e.BasicTest;
+import pl.cyfronet.s4e.TestDbHelper;
+import pl.cyfronet.s4e.bean.AppUser;
+import pl.cyfronet.s4e.bean.PasswordReset;
+import pl.cyfronet.s4e.controller.request.PasswordChangeRequest;
+import pl.cyfronet.s4e.controller.request.PasswordResetRequest;
+import pl.cyfronet.s4e.data.repository.AppUserRepository;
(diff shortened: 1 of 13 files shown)


s4e-backend/src/test/java/pl/cyfronet/s4e/service/PasswordServiceTest.java
```diff
@@ -0,0 +1,246 @@
...
+                .expiryTimestamp(LocalDateTime.now().plus(EXPIRE_IN))
+                .build();
+        passwordResetRepository.save(token);
+        try {
+            passwordService.validate(token.getToken());
+        } catch (NotFoundException e) {
+            fail("Token not found");
+        } catch (PasswordResetTokenExpiredException e) {
+            fail("Token expired");
+        }
```
jswk: Remove the try-catch here. If it throws, the test will fail anyway. Alternatively, pass the exception as the second argument of the `fail` call.

s4e-backend/src/test/java/pl/cyfronet/s4e/service/PasswordServiceTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertThat(passwordService.findByEmail(PROFILE_EMAIL), isEmpty());
+        passwordService.createPasswordResetTokenForUser( PROFILE_EMAIL);
+        assertThat(passwordService.findByEmail(PROFILE_EMAIL), isPresent());
+    }
+
+    @Test
+    public void shouldntCreatePasswordResetTokenForUser() {
+        assertThat(passwordService.findByEmail(PROFILE_EMAIL), isEmpty());
+        assertThrows(NotFoundException.class, () -> passwordService.createPasswordResetTokenForUser( "mail@email.pl"));
+        assertThat(passwordService.findByEmail(PROFILE_EMAIL), isEmpty());
```
jswk: You should assert that `findByEmail("mail@email.pl")` return empty, not `findByEmail(PROFILE_EMAIL)`.

s4e-backend/src/test/java/pl/cyfronet/s4e/service/PasswordServiceTest.java
kmarszalek: yeah, not sure how that happen :D

s4e-backend/src/test/java/pl/cyfronet/s4e/service/PasswordServiceTest.java
kmarszalek: nice catch :)

Issue #371, opened 2020-02-25
Rework Product/Scene schema to fit GS-IM


jswk:
See #340 for more discussion.

- [x] Update GeoServer and change S3Geotiff endpoint to `mailto` #374 
- [x] Change postgresql version to a version with postgis and update the db to the latest version #375 
- [ ] Actually rework the schema and make it work with GS #376

PR #376, opened 2020-03-03
Modify Product and Scene to match with GeoServer ImageMosaic

### LINKED ISSUE #389: Change front to work with new GS ImageMosaic
https://github.com/cyfronet-fid/sat4envi/pull/376

- Requests to WMS with extra parameter "time" [Scene->timestamp]
- Layer name is now in Product->layerName field

jswk:
Move the layer_name from Scene to Product, as we will have a single
layer per Product after this change.
Drop Scene.created, as it won't apply anymore.
Add granule_path and footprint to Scene, which are required by IM.

There is a lot of fixes in tests, as the newly added fields are required
to be non-null.

The WebhookController no longer communicates directly with GS.

The SeedProducts now only creates views named `scene_<product_name>`.
**The stores and layers are left not-configured for now, the
implementation should be put in GeoServerService::addLayer.**
A scene is not created if the corresponding GeoTiff doesn't exist in S3,
the process will fail if the S3 connection is misconfigured.

Part of #371.
Closes #389.


kmarszalek:
i will modify commit msg. but later :)

s4e-backend/src/main/java/pl/cyfronet/s4e/service/GeoServerService.java
```diff
@@ -31,8 +32,25 @@ public void resetWorkspace() {
         geoServerOperations.createWorkspace(geoServerProperties.getWorkspace());
     }
 
-    public void addLayer(Product product) {
-        // FIXME: implement
+    @Transactional(rollbackFor = RestClientResponseException.class)
+    public void addStoreAndLayer(Product product) {
+        // Both the coverage store, coverage and layer (the last one with workspace prefix) names will be the same
+        String gsName = product.getName().toLowerCase();
```
jswk: The name of the GS layer should be product.getLayerName()

s4e-backend/src/main/java/pl/cyfronet/s4e/geoserver/op/GeoServerOperations.java
```diff
@@ -156,11 +197,36 @@ public void deleteCoverageStore(String workspace, String coverageStore, boolean
...
+                "   \"name\": " + coverage + ",\n" +
+                "   \"title\":" + coverage + ",\n" +
+                "   \"enabled\":true,\n" +
+                "   \"type\": \"raster\",\n" +
+                "   \"metadata\":\n{\n" +
+                "       \"entry\":[{\n" +
+                "           \"@key\":\"time\",\n" +
+                "           \"dimensionInfo\":{\n" +
+                "               \"enabled\":true,\n" +
+                "               \"presentation\":\"LIST\",\n" +
```
jswk: Better to use the presentation type with duration, or sth like that, so that not all granules are listed.

s4e-backend/src/main/java/pl/cyfronet/s4e/geoserver/op/GeoServerOperations.java
```diff
@@ -156,11 +197,36 @@ public void deleteCoverageStore(String workspace, String coverageStore, boolean
     }
 
     public void createS3Coverage(String workspace, String coverageStore, String coverage) {
-        URI url = UriComponentsBuilder.fromHttpUrl(
-                geoServerProperties.getBaseUrl() + "/workspaces/{workspace}/coveragestores/{coveragestore}/coverages")
-                .buildAndExpand(workspace, coverageStore).toUri();
-        val entity = httpEntity(new CreateS3CoverageRequest(workspace, coverageStore, coverage));
-        restTemplate().postForObject(url, entity, String.class);
+        String body = "{\n" +
```
jswk: Wouldn't it be better to construct it the way other payloads in this class are?

s4e-backend/src/main/java/pl/cyfronet/s4e/geoserver/sync/GeoServerSynchronizer.java
```diff
@@ -18,14 +19,25 @@
...
+                geoServerService.addStoreAndLayer(product);
+            }
+        }
+    }
+
     @Transactional
     public void synchronizeOverlays() {
         log.info("Creating styles");
-        for (val sldStyle: sldStyleRepository.findAll()) {
+        for (val sldStyle : sldStyleRepository.findAll()) {
```
jswk: Why break formatting here?

s4e-backend/src/test/java/pl/cyfronet/s4e/geoserver/op/SeedProductsTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+@Profile({"test"})
```
jswk: You don't need the curly braces here, also, shouldn't it be "integration" instead of "test"?

s4e-backend/src/test/resources/geoserver/datastore.properties
```diff
@@ -0,0 +1,5 @@
+SPI=org.geotools.data.postgis.PostgisNGJNDIDataStoreFactory
```
jswk: Do you need these resources here? Wouldn't it be enough to use the ones from the main resources? Aren't they the same?

README.md
```diff
@@ -145,12 +146,18 @@ Then, run the backend with properties:
 spring.profiles.active=development
 seed.products.data-set=s4e-demo
 s3.bucket=s4e-demo
+s3.access-key=<access_key>
+s3.secret-key=<secret_key>
+S3_ENDPOINT=<path to storage>
```
jswk: This should read "s3.endpoint" instead of uppercased version

s4e-backend/src/main/java/pl/cyfronet/s4e/geoserver/sync/GeoServerSynchronizer.java
kmarszalek: for this type its default formatting, there are multiple examples in our code, commited also by you :)
for me this way is cleaner to read. also it was me doing ctrl+alt+l, not explicity changing that line.

s4e-backend/src/main/java/pl/cyfronet/s4e/geoserver/op/request/CreateS3CoverageRequest.java
```diff
@@ -29,18 +47,57 @@
         String name;
         String title;
         String srs = RequestConstants.SRS;
+        String type = RequestConstants.S3_COVERAGE_TYPE;
```
jswk: It isn't reflected in the javadoc above

s4e-backend/src/main/java/pl/cyfronet/s4e/geoserver/op/request/CreateS3CoverageRequest.java
```diff
@@ -29,18 +47,57 @@
...
+        boolean enabled = true;
+        String presentation = "CONTINUOUS_INTERVAL";
+        String units = "ISO8601";
+        DefaultValue defaultValue;
+    }
+
+    @Value
+    @Builder
+    private static class DefaultValue {
+        String strategy = "MINIMUM";
```
jswk: I thought about it, and it is fine to be MIINIMUM, even though it could come to mind that MAXIMUM is a better strategy. It could be because max means serving the most recent granule - however, that would open up a possibility to get unlicensed EUMETSAT granules, so better make asking for them explicit - and easier to catch in the gateway.
Anyway, because of that add a comment about this, because we (or sb else) may later forget this reasoning and fall for the licensing problem.

s4e-backend/src/main/java/pl/cyfronet/s4e/geoserver/op/GeoServerOperations.java
```diff
@@ -123,12 +129,48 @@ public void createExternalShpDataStore(String workspace, String dataStoreName, S
...
+        ClassLoader classloader = Thread.currentThread().getContextClassLoader();
+        String mosaic = "Name=" + coverageStore + "\n" +
+                "TypeName=scene_" + coverageStore + "\n" +
+                "Levels=5000,5000\n" +
+                "LevelsNum=1\n";
+        InputStream streamMosaic = new SequenceInputStream(new ByteArrayInputStream(mosaic.getBytes()), classloader.getResourceAsStream("geoserver/mosaic.properties"));
+        InputStream streamStore = classloader.getResourceAsStream("geoserver/datastore.properties");
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        try {
+            ZipOutputStream zos = new ZipOutputStream(bos);
```
jswk: You should use try-with-resources here for all the streams apart from bos, otherwise there may be memory leaks in case of errors.

s4e-backend/src/main/java/pl/cyfronet/s4e/geoserver/op/GeoServerOperations.java
```diff
@@ -123,12 +129,48 @@ public void createExternalShpDataStore(String workspace, String dataStoreName, S
...
+            zos.putNextEntry(new ZipEntry("datastore.properties"));
+            zos.write(streamStore.readAllBytes());
+            zos.closeEntry();
+            streamMosaic.close();
+            streamStore.close();
+            zos.close();
+        } catch (FileNotFoundException ex) {
+            ex.printStackTrace();
+        } catch (IOException ex) {
+            ex.printStackTrace();
```
jswk: log?

s4e-backend/src/main/java/pl/cyfronet/s4e/service/GeoServerService.java
```diff
@@ -31,8 +32,25 @@ public void resetWorkspace() {
         geoServerOperations.createWorkspace(geoServerProperties.getWorkspace());
     }
 
-    public void addLayer(Product product) {
-        // FIXME: implement
+    @Transactional(rollbackFor = RestClientResponseException.class)
+    public void addStoreAndLayer(Product product) {
+        // Both the coverage store, coverage and layer (the last one with workspace prefix) names will be the same
+        String gsName = product.getLayerName().toLowerCase();
```
jswk: Do the lowercase thing when populating the field, and add it as a regex constrain on the field.

s4e-backend/src/test/java/pl/cyfronet/s4e/geoserver/op/SeedProductsTest.java
```diff
@@ -0,0 +1,180 @@
...
+            if ((i + 1) % 100 == 0) {
+                log.info((i + 1) + "/" + count + " scenes of product '" + product.getName() + "' processed");
+            }
+        }
+    }
+
+    private void createViews(List<Product> products) {
+        for (val product: products) {
+            String id = product.getId().toString();
+            String name = product.getName().toLowerCase();
```
jswk: I'd use layerName here. The same in tests

michal-szostak:
I'm taking over this PR, front end changes will follow.

Issue #401, opened 2020-04-03
Provide seeds for products and layers for e2e


michal-szostak:
Currently E2E tests can not test much, as the products and overlayers are not seeded.
This should be fixed. @jswk should provide further information what exactly should be provided.
At the end e2e tests for product navigation should be uncommented, and they will be the validation of this issue.

jswk:
@wziajka created docker image fiddev/minio-test:1.0.0, which is properly seeded for integration and E2E tests.

I will use it in docker-compose.yml to provide environment for the tests.
Solving this will possibly also close #80 and #260.

PR #413, opened 2020-04-07
[#401] Run e2e and integration tests in gh-actions, modify dc files

jswk:
The docker-compose file is split into two files: docker-compose.yml and
docker-compose-test.yml. The former contains containers which are running the
application, while the latter contains containers used for tests.

Different containers are required for different types of tests.
Backend unit-tests: db-test.
Backend integration-tests: db-test, minio-test, geoserver-test.
Front e2e tests: db-test, minio-test, geoserver-test, s4e-web-e2e and
s4e-backend-e2e.

@michal-szostak: Check if I modified the e2e tests correctly. Have a look at the README too.

@wziajka: See if the modified docker-compose files make sense.

Closes: #401.


s4e-web/cypress.json
```diff
@@ -1,5 +1,5 @@
 {
-  "baseUrl": "http://localhost:4200",
+  "baseUrl": "http://localhost:4202",
```
michal-szostak, line 2: Why change this port?

Issue #420, opened 2020-04-15
Implement favourite products in front-end


jswk:
### Previews

https://projects.invisionapp.com/share/S9RJNE8WXUB#/screens/413514590

### Endpoints

`PUT /api/products/{product-id}/favourite`
sets product to be favourite

`DELETE /api/products/{product-id}/favourite`
removes favourite status

GET `/api/products/`
every product now has additional field `favourite:boolean`

Blocked by: #404

PR #424, opened 2020-04-16
[#420] Favorites products

danielkryska:


s4e-web/src/app/views/map-view/state/overlay/overlay.factory.spec.ts
```diff
@@ -4,5 +4,6 @@ import {Overlay} from './overlay.model';
 export const OverlayFactory = Factory.makeFactory<Overlay>({
   id: Factory.each(i => `overlay:${i}`),
   caption: Factory.each(i => `Overlay #${i}`),
-  type: 'wms'
+  type: 'wms',
+  favourite: Factory.each(i => !!Math.round(Math.random()))
```
michal-szostak, line 8: This is generally not a good idea, as it's rather useless in tests (non deterministic input). In most cases if we do not expect fields to be unique settling for static value is a better idea. Eg. `false` in this case.

s4e-web/src/app/views/map-view/view-manager/layer-picker/layer-picker.component.html
```diff
@@ -6,14 +6,19 @@
...
+      <figure class="list__icon" (click)="itemSelected.emit(item.cid)">
         <img src="../../../../../assets/images/ico_product_default.png" alt="Default product icon" i18n-alt />
       </figure>
-      <button class="list__name">
+      <button class="list__name" (click)="itemSelected.emit(item.cid)">
         {{item.caption}}
       </button>
-      <button class="icon__visibility">
+      <button class="icon__visibility" (click)="itemSelected.emit(item.cid)">
         <img src="../../../../../assets/images/ico_eye_open.svg" alt="Eye icon - product visibility" i18n-alt />
```
michal-szostak, line 22: We're generally using Font Awesome in the project. I do recall that stars are part of the FA library. Check for FA usage in the project. (otherwise this is the way assets should be included, but in this particular case it might not be necessary)

s4e-web/src/app/views/map-view/view-manager/layer-picker/layer-picker.component.ts
```diff
@@ -21,4 +21,6 @@ export class LayerPicker implements OnInit {
 
   ngOnInit() {
   }
+
```
michal-szostak, line 24: I would rather suggest adding

`@Input() showFavouriteButtons: boolean = false;`
 as an input.

s4e-web/src/app/views/map-view/state/common.model.ts
```diff
@@ -4,4 +4,5 @@ export interface IUILayer {
   cid: ID;
   caption: string;
   active: boolean;
+  favourite?: boolean;
```
michal-szostak, line 7: Settling for field which can not be undefined is a better option. And just setting it always to false in `overlay.model.ts`

```
export function convertToUIOverlay(overlay: Overlay, geoServerUrl: string, active: boolean = false): UIOverlay {
  return {
    ...overlay,
    olLayer: new Image({
      source: new ImageWMS({
        crossOrigin: 'Anonymous',
        url: geoServerUrl,
        serverType: 'geoserver',
        params: {LAYERS: overlay.id}
      })
    }),
    cid: overlay.id,
    active: active,
    favourite: false
  };
}
```

s4e-web/src/app/views/map-view/state/overlay/overlay.model.ts
```diff
@@ -14,6 +14,7 @@ export interface Overlay {
   id: string;
   type: OverlayType;
   caption: string;
+  favourite?: boolean;
```
michal-szostak, line 17: Overlays do not have this field

jswk:
I'm also going to mention to remember the changelog. Have a look at CHANGELOG.md, and create an entry there.

s4e-web/src/app/views/map-view/state/overlay/overlay.factory.spec.ts
danielkryska: ok

s4e-web/src/app/views/map-view/state/common.model.ts
danielkryska: It's fine with me as it's our use case, although in yesterday conversation you mention that favourite will not be included in all converted items. Example Products can be favorite, but layers won't be

s4e-web/src/app/views/map-view/state/overlay/overlay.model.ts
danielkryska: This will cause incompatibility when some of its children will  inherit also by UILayer

s4e-web/src/app/views/map-view/view-manager/layer-picker/layer-picker.component.html
danielkryska: I was following your code style, that is written beneath lines of mine. In that case, I'll change it.

s4e-web/src/app/views/map-view/view-manager/layer-picker/layer-picker.component.ts
danielkryska: I think that logic is simpler if we check if the object contains the favorite field instead of adding new Inputs.

kmarszalek:
please use "Draft PR"  for WIP pull-request :)

s4e-web/src/app/views/map-view/state/product/product.service.ts
```diff
@@ -57,6 +57,35 @@ export class ProductService {
 
   }
 
+  toggleFavorite(ID: number, isFavourite: boolean) {
+    if (isFavourite) {
+      console.log(ID, isFavourite, `${this.CONFIG.apiPrefixV1}/products/${ID}/favourite`);
```
kmarszalek, line 62: log :)

kmarszalek:
i made myself a reviewer.. not sure when and what for but..
- which we should use US or GB? -> Favorite vs Favourite :)
- add "s" in my changelog entry (thanks) :)
- still got console.log, why?

danielkryska:
**How it looks like?**

image

Issue #423, opened 2020-04-16
Notifications module


michal-szostak:
Create notification module which will allow to post notifications to user interface.

Notifications should be able to show errors, warnings, success and info. 
They should have automatic dismiss time, or be dismissable on click.
Automatic dismiss should be reinstantiated after store reload, for development purposes.
Notification components should be extensible for the future, but provide basic component which can display every basic notification type.

This issue may or may not include styling, depending on @abacz time allocation.

PR #431, opened 2020-04-26
[#423] Notifications module

michal-szostak:
## What is in this PR?

Notification module with basic functionality and very basic (based on bootstrap) styles.
@abacz is assumed to style them further, but for now they should be at least presentable.

## How does it looks?

image

## Other noteworthy things

The module is a separate library, which cuts it's dependencies to the main application.

Closes: #423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ The most recent changes are on top, in each type of changes category.
 
 - [Bugfix] places input should keep state [#427](https://github.com/cyfronet-fid/sat4envi/issues/427)
 
+### Added
 
+- Notifications module [#423](https://github.com/cyfronet-fid/sat4envi/issues/423)
 
 ## [v5.1.0]
 
--- a/s4e-web/package.json
+++ b/s4e-web/package.json
@@ -3,11 +3,19 @@
   "version": "0.0.2",
   "scripts": {
     "ng": "ng",
+    "start-notifications": "ng build notifications --watch",
+    "build-libraries": "ng build notifications",
+    "prestart": "node_modules/.bin/npm run build-libraries",
     "start": "ng serve --proxy-config proxy.conf.json --configuration hmr",
+    "prestart-private": "node_modules/.bin/npm run build-libraries",
     "start-private": "ng serve --proxy-config proxy.conf.json --configuration hmr.private",
+    "prestart-production": "node_modules/.bin/npm run build-libraries",
     "start-production": "ng serve --proxy-config proxy.conf.json --configuration production",
+    "prestart-production-private": "node_modules/.bin/npm run build-libraries",
     "start-production-private": "ng serve --proxy-config proxy.conf.json --configuration production.private",
+    "prebuild": "node_modules/.bin/npm run build-libraries",
     "build": "ng build --prod",
+    "prebuild-private": "node_modules/.bin/npm run build-libraries",
     "build-private": "ng build --configuration production.private",
     "test": "ng test",
     "lint": "ng lint",
@@ -76,6 +84,7 @@
   "devDependencies": {
     "@angular-builders/jest": "^7.3.1",
     "@angular-devkit/build-angular": "^0.8.9",
+    "@angular-devkit/build-ng-packagr": "~0.10.0",
     "@angular/cli": "^7.0.7",
     "@angular/compiler-cli": "^7.2.4",
     "@angular/language-service": "^7.2.4",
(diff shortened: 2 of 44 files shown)


s4e-web/projects/notifications/src/lib/components/general-notification/general-notification.component.html
```diff
@@ -0,0 +1,10 @@
+<div [ngClass]="{clickable: notification.clearable}"
+     (click)="activated.emit()"
+     class="notification-container ui-notification {{notification.type}}"
+     role="alert">
+<!--  <h3 *ngIf="notification.caption">{{notification.caption}}</h3>-->
+  <div class="message" [innerHtml]="notification.content"></div>
+<!--  <div *ngIf="notification.clearable" class="notification-action">-->
```
danielkryska, line 7: Do you need those comments?

s4e-web/projects/notifications/src/lib/state/notification.service.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [NotificationsModule.forRoot()],
+    });
+
+    notificationService = TestBed.get(NotificationService);
+    notificationStore = TestBed.get(NotificationStore);
+  });
+
+  it('TODO - redo tests', () => {
```
danielkryska, line 19: What test case is it?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The most recent changes are on top, in each type of changes category.
 - Favorites products [#420] (https://github.com/cyfronet-fid/sat4envi/issues/420)
 - Favourite Products [#403](https://github.com/cyfronet-fid/sat4envi/issues/403)
 - Forms module [#370](https://github.com/cyfronet-fid/sat4envi/issues/370)
+- Notifications module [#423](https://github.com/cyfronet-fid/sat4envi/issues/423)
```
jswk, line 23: This should go to the unreleased section

s4e-web/package.json
```diff
@@ -3,11 +3,19 @@
   "version": "0.0.2",
   "scripts": {
     "ng": "ng",
+    "start-notifications": "ng build notifications --watch",
+    "build-libraries": "ng build notifications",
+    "prestart": "node_modules/.bin/npm run build-libraries",
```
jswk, line 8: Does it have to be repeated so many times? Isn't there some way to automatically run the build-libraries when running/building the app?

s4e-web/projects/notifications/src/lib/components/notification-outlet/notification-outlet.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(private notificationQuery: NotificationQuery,
+              @Inject(NotificationAction) private notificationActions: any[],
+              private notificationService: NotificationService) { }
+
+  ngOnInit() {
+    this.notifications$ = this.notificationQuery.selectAll();
+  }
+
+  activated(id: ID, ctx?: string) {
+    console.log(this.notificationActions);
```
jswk, line 40: Should this be logged?

michal-szostak: right, my bad

s4e-web/package.json
michal-szostak: unfortunately there is no way (at least I could not found one) that was one of the "issues" of using library which I mentioned. They are not included in the standard `ng build`

michal-szostak:
> I have a general question, why there are two types of NotificationComponent: DynamicNC and GeneralNC. Dynamic seems to be used in the app, while I don't see the binding to the General. However, I'm probably missing something. Could you explain this architecture?

Sure. The `DynamicNC` is a dynamic component which instantiates concrete notification component based on passed `notification` input.

The coupling is loose on purpose and mappings between types and components is done in providers (see `notifications.module.ts` there are providers like: `{provide: NotificationClazz, useValue: {name: 'GeneralNotificationComponent', component: GeneralNotificationComponent}, multi: true},`)

This kind of architecture allows extending notifications (providing custom notifications with custom data) outside of the library. For example one could create custom notification component which shows progress bar of loading some resources, or something.

s4e-web/package.json
jswk: Understood, thanks for clarification

s4e-web/projects/notifications/src/lib/components/general-notification/general-notification.component.html
michal-szostak: They are for @abacz . I opted out of using caption and extended elements, but this is just my humble opinion. Maybe @abacz will decide that notifications should have captions, etc. That's why I left it here, as a hint.

s4e-web/projects/notifications/src/lib/state/notification.service.spec.ts
michal-szostak: This is the test case which had to be done (it should be good now)

Issue #430, opened 2020-04-22
Global error handling for http requests


bwilk:
- Global error handling using interceptor
  https://scotch.io/bar-talk/error-handling-with-angular-6-tips-and-best-practices192
  https://www.tektutorialshub.com/angular/angular-http-error-handling/
  - [x] error 401: Unathorized/error 403: forbidden
    - redirect to login page
    - [x] after login backward to last url (back linking) using additional param in url
          ?back='/path'
    - notyfication
  - [x] error 400: bad request/error 101: timeout
    - notification
  - [x] error 404: not found
    - redirect to error page
    - use specific for thi use case message
  - [x] error 500: internal server error/error 502: Bad gateway
    - redirect to error page
    - use specific for thi use case message
  - [x] mock errors in e2e
        https://docs.cypress.io/api/commands/server.html#Options
  - [x] mock errors in Unit Tests
  - [x] use notifications in TODO places
  - [x] provide Accepted_Error_Codes, Skiped_Error_Codes Headers
        https://stackoverflow.com/questions/46469349/how-to-make-an-angular-module-to-ignore-http-interceptor-added-in-a-core-module?answertab=votes#tab-top
  - [x] update service forms header (search by http.post)
  - [ ]  ~~RXJS operator (e.g. catchStoreAndThrow(store))~~

bwilk:
@kmarszalek is contact point for the API functionalities

PR #432, opened 2020-04-27
[#430] global error handling for http requests

danielkryska:
Global error handling, catch beneath errors. Each of them is handled differently

ERROR_CODE_UNAUTHORIZED = 401;
ERROR_CODE_FORBIDDEN = 403;

Should navigate to `/login` page with notification:
`Wystąpił błąd: Nie jesteś zalogowany lub twoja sesja się przedawniła`

---
ERROR_CODE_BAD_REQUEST = 400;
ERROR_CODE_TIMEOUT = 101;

Should show notification with message:
`Wystąpił błąd: Nieprawidłowe zapytanie lub zbyt długi czas oczekiwania na odpowiedź serwera`

---
ERROR_CODE_NOT_FOUND = 404;

Should navigate to `/errors/404` page

---
ERROR_CODE_INTERNAL_SERVER_ERROR = 500;
ERROR_CODE_BAD_GATEWAY = 502;

Should navigate to `/errors/500` or `/errors/502` page, with a possibility to return to the last page (that was visited)

---

**UPDATE** (@michal-szostak): There are some exceptions to the rule for POSTS in most places (in all forms) there is no default notification for error `400` as it is handled by the form itself.

For manual testing I recommend this chrome extension: https://chrome.google.com/webstore/detail/interceptor/enenfaicdcfgcnjmiigcjbmlbaoapnen?hl=en

It is not perfect but it can mock ajax responses in the browser, which enables easy error checking without having to mess with the server setup.

### DIFF
--- a/s4e-web/src/app/errors/errors.component.spec.ts
+++ b/s4e-web/src/app/errors/errors.component.spec.ts
@@ -1,0 +1,53 @@
+import { By } from '@angular/platform-browser';
+import { ParamMap } from '@angular/router/src/shared';
+import { RouterTestingModule } from '@angular/router/testing';
+import { HttpClientModule } from '@angular/common/http';
+import { ActivatedRoute, convertToParamMap } from '@angular/router';
+import { ErrorsModule } from './errors.module';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { ErrorsComponent } from './errors.component';
+import { ReplaySubject, Observable, Subject } from 'rxjs';
+
+class ActivatedRouteStub {
+  paramMap: Subject<ParamMap> = new ReplaySubject(1);
+  queryParamMap: Subject<ParamMap> = new ReplaySubject(1);
+
+  constructor() {
+    this.paramMap.next(convertToParamMap({}));
+    this.queryParamMap.next(convertToParamMap({backLink: '/'}));
+  }
+}
+
+
+describe('ErrorComponent', () => {
+  let component: ErrorsComponent;
+  let fixture: ComponentFixture<ErrorsComponent>;
+  let route: ActivatedRouteStub;
+  const backLink = '/back-link';
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [ ErrorsModule, HttpClientModule, RouterTestingModule ],
+      providers: [{provide: ActivatedRoute, useClass: ActivatedRouteStub}]
+    })
+    .compileComponents();
+    route = TestBed.get(ActivatedRoute);
+  }));
+
+  beforeEach(() => {
+    route.paramMap.next(convertToParamMap({errorCode: '404'}));
+    route.queryParamMap.next(convertToParamMap({backLink}));
+    fixture = TestBed.createComponent(ErrorsComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should display 404 error', () => {
+    expect(fixture.debugElement.query(By.css('.title')).nativeElement.textContent).toBe('Błąd 404: nie znaleziono poszukiwanej strony')
+  })
+});
--- a/s4e-web/src/app/errors/errors.component.ts
+++ b/s4e-web/src/app/errors/errors.component.ts
@@ -1,0 +1,61 @@
+import {untilDestroyed} from 'ngx-take-until-destroy';
+import {ActivatedRoute} from '@angular/router';
+import {Component, OnDestroy, OnInit} from '@angular/core';
+import {HTTP_404_NOT_FOUND, HTTP_500_INTERNAL_SERVER_ERROR, HTTP_502_BAD_GATEWAY} from './errors.model';
+
+@Component({
+  selector: 's4e-errors',
+  templateUrl: './errors.component.html',
+  styleUrls: ['./errors.component.scss']
+})
+export class ErrorsComponent implements OnInit, OnDestroy {
+  public actualError = ErrorsComponent._getErrorBy(null);
+  public back_link?: string;
+
+  constructor(private _activatedRoute: ActivatedRoute) {
+  }
+
+  protected static _getErrorBy(code: number | null) {
+    switch (code) {
+      case HTTP_404_NOT_FOUND:
+        return {
+          title: 'Błąd 404: nie znaleziono poszukiwanej strony',
+          description: 'Strona, której szukasz nie istnieje, bądź została przeniesiona'
+        };
+      case HTTP_502_BAD_GATEWAY:
+        return {
+          title: 'Błąd 502: brak odpowiedzi serwera lub błędna odpowiedź',
+          description: 'Sprawdź swoje ustawienia przegladarki lub skontaktuj się z administratorem'
+        };
+      case HTTP_500_INTERNAL_SERVER_ERROR:
+        return {
+          title: 'Błąd 500: błąd po stronie serwera',
+          description: 'Skontaktuj się z administratorem w przypadku powtarzającej się awarii'
+        };
+      case null:
+      default:
+        return {
+          title: 'Wystąpił nieznany błąd',
+          description: 'Skontaktuj się z administratorem w przypadku powtarzającej się awarii'
+        };
+    }
+  }
+
+  ngOnInit() {
+    this._activatedRoute.paramMap
+      .pipe(untilDestroyed(this))
+      .subscribe((params) => {
+        const actualErrorCode = +params.get('errorCode');
+        this.actualError = ErrorsComponent._getErrorBy(actualErrorCode);
+      });
+
+    this._activatedRoute.queryParamMap
+      .pipe(untilDestroyed(this))
+      .subscribe((params) => {
+        this.back_link = params.get('back_link');
+      });
+  }
+
+  ngOnDestroy() {
+  }
+}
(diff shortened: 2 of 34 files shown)


s4e-web/src/app/utils/error-interceptor/error.interceptor.ts
```diff
@@ -2,23 +2,85 @@ import { Injectable } from '@angular/core';
...
+    return next.handle(request)
+      .pipe(
+        catchError((error: HttpErrorResponse) => {
+          if (HttpErrorHelper.hasSkipHeader(request)) {
+            return HttpErrorHelper.skipedHttpEvent(request, next);
+          }
+
+          if (HttpErrorHelper.isClientSideError(error)) {
+            console.error('Error Event');
+            // TODO How to handle this error?
```
michal-szostak, line 47: We should pass it. I think it can only happen if there is another interceptor earlier in the chain, and it throws error.

s4e-web/src/app/views/settings/groups/state/group.service.ts
```diff
@@ -23,7 +23,13 @@ export class GroupService {
     this.store.setLoading(true);
     this.store.setError(null);
     this.http.get<IPageableResponse<Institution>>(`${this.config.apiPrefixV1}/institutions/${institutionSlug}/groups`)
-      .pipe(finalize(() => this.store.setLoading(false)))
+      .pipe(
+        finalize(() => this.store.setLoading(false)),
+        catchError(error => {
+          this.store.setError(error);
+          return throwError(error);
+        })
```
michal-szostak, line 31: This happens very often now, I think that you can consider making our own RXJS operator (e.g. `catchStoreAndThrow(store)`)

s4e-web/src/app/utils/error-interceptor/error.interceptor.ts
danielkryska: ok

jswk:
Ah, forgot about my usual comment about changelog. ;)

s4e-web/src/app/utils/error-interceptor/error.interceptor.ts
```diff
@@ -2,23 +2,85 @@ import { Injectable } from '@angular/core';
...
+import { Router } from '@angular/router';
+
+const ERROR_CODE_UNATHORIZED = 401;
+const ERROR_CODE_FORBIDEN = 403;
+const ERROR_CODE_BAD_REQUEST = 400;
+const ERROR_CODE_TIMEOUT = 101;
+
+export const ERROR_CODE_NOT_FOUND = 404;
+export const ERROR_CODE_INTERNAL_SERVER_ERROR = 500;
+export const ERROR_CODE_BAD_GATEWAY = 502;
```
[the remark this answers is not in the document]
danielkryska: I  don't found any, those are available only in digital form

s4e-web/src/app/views/settings/groups/state/group.service.ts
```diff
@@ -36,11 +42,7 @@ export class GroupService {
       ...value,
       membersEmails: value.membersEmails._
     }).pipe(
-      finalize(() => this.store.setLoading(false)),
-      catchError(error => {
-        this.store.setError(error);
-        return throwError(error);
-      })
```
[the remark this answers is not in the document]
danielkryska: My bad, I was trying to do more generic error handle of storages, but It's impossible. Changes in catching errors weren't stashed before committing. I will fix those.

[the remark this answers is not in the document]
jswk: Great, thanks!

s4e-web/src/app/views/settings/groups/state/group.service.ts
danielkryska: I think it's better to consider such changes during the refactoring session. For now, the whole feature is big enough to be considered as the only thing to be done.

s4e-web/.angulardoc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "repoId": "0798367e-3e97-486a-a4af-60d3997da6a1",
+  "lastSync": 0
+}
```
michal-szostak, line 4: I'm guessing that this is VS file, should it be included in the repository?

s4e-web/src/app/views/login/login.component.ts
```diff
@@ -15,12 +15,19 @@ import {GenericFormComponent} from '../../utils/miscellaneous/generic-form.compo
...
               private sessionService: SessionService,
-              private sessionQuery: SessionQuery) {
+              private sessionQuery: SessionQuery,
+              private _activatedRoute: ActivatedRoute
+  ) {
     super(fm, router, sessionQuery, 'login');
+
+    this._activatedRoute.queryParamMap
+      .subscribe((queryParams) => {
+        this.sessionService.backLink = queryParams.get('backLink');
```
michal-szostak, line 29: This is super minor, but our convention for query params is `snake_case`, so it's preferable for `queryParams.get('backLink')` to be `queryParams.get('back_link')`

s4e-web/src/app/views/map-view/state/product/product.service.ts
```diff
@@ -110,7 +109,13 @@ export class ProductService {
 
     this.http.get<string[]>(`${this.CONFIG.apiPrefixV1}/products/${activeProduct.id}/scenes/available`,
       {params: {tz: this.CONFIG.timezone, yearMonth: dateF}})
-      .pipe(finalize(() => this.store.setLoading(false)))
+      .pipe(
+        finalize(() => this.store.setLoading(false)),
+        catchError(error => {
+          this.store.setError(error);
+          return throwError(error);
+        }),
```
michal-szostak, line 117: As I mentioned before, this piece of code repeats very often, so it should be extracted into separate rxjs operator

s4e-web/src/app/views/settings/state/institution.service.ts
```diff
@@ -28,7 +28,13 @@ export class InstitutionService {
     this.store.setLoading(true);
     this.store.setError(null);
     this.http.get<IPageableResponse<Institution>>(`${this.s4EConfig.apiPrefixV1}/institutions`)
-      .pipe(finalize(() => this.store.setLoading(false)))
+      .pipe(
+        finalize(() => this.store.setLoading(false)),
+        catchError(error => {
+          this.store.setError(error);
+          return throwError(error);
+        }),
```
michal-szostak, line 36: there is error handling in `.subscribe(...)` why replicate it here?

s4e-web/src/app/errors/errors.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+import { ErrorsComponent } from './errors.component';
+
+describe('ErrorComponent', () => {
+  let component: ErrorsComponent;
+  let fixture: ComponentFixture<ErrorsComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ ErrorsComponent ]
+    })
```
michal-szostak, line 12: It's better to import whole module which contains component. This way you make sure, that the module itself has all the dependencies for the component. In this case this is rather trivial, but that's the general rule. And it makes sure that as you add dependencies to the module your tests won't *desynchronize*

s4e-web/src/app/errors/errors.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  public actualError;
+  public backLink?: string;
+
+  constructor(private _activatedRoute: ActivatedRoute) {
+    this._activatedRoute.paramMap.subscribe((params) => {
+      this.backLink = params.get('backLink');
+
+      const actualErrorCode = +params.get('errorCode');
+      this.actualError = ErrorsComponent._getErrorBy(actualErrorCode);
+    });
```
michal-szostak, line 20: All subscribes, etc should be done in `ngOnInit` (implemet interface `OnInit` and `OnDestroy`.
Additionally do not forget to unsubscribe from all subscriptions. The easiest way to do this in the context of the component is to have it implement `OnDestroy` and use `.pipe(untilDestoryed(this)).subscribe(...)`
Check other components which subscribe to things.

s4e-web/src/app/utils/error-interceptor/error.interceptor.ts
```diff
@@ -1,24 +1,99 @@
...
+
+  private _handleHttpErrorResponses = (error: HttpErrorResponse) => {
+    console.log(`error status : ${error.status} ${error.statusText}`);
+    switch (error.status) {
+      case ERROR_CODE_FORBIDDEN:
+      case ERROR_CODE_UNAUTHORIZED:
+        this._router.navigate(
+          ['/login'],
+          {
+            queryParams: {backLink: this._backLink}
```
michal-szostak, line 78: when reading and creating back links later you add `/` prefix. I think it would be better to add it here.

s4e-web/src/app/utils/error-interceptor/error.interceptor.ts
```diff
@@ -1,24 +1,99 @@
...
+          }
+        );
+        // TODO @Michał Add notification for user
+        break;
+      case ERROR_CODE_TIMEOUT:
+      case ERROR_CODE_BAD_REQUEST:
+        // TODO @Michał Add notification for user
```
michal-szostak, line 85: Notifications are available in the master, so you can add them here.

s4e-web/.angulardoc.json
danielkryska: It shouldn't, good to know that it was pushed...

s4e-web/src/app/errors/errors.component.spec.ts
danielkryska: It has been generated automatically...

s4e-web/src/app/errors/errors.component.ts
danielkryska: Next thing to be added to refactoring stuff, I'll update refactoring list with it. There is also another type of solution. Http Observable can be transformed into Promise, so there will be any need of subscription (just solved by using then()).

PR #442, opened 2020-04-30
Update geoserver image to fiddev/geoserver:1.4.0-GS2.17.0

jswk:
Apart from the GeoServer version update, I changed the way of setting S3 credentials: now they are specified as envvars for the container.

@wojt: How do you feel about this way of setting S3 credentials?

@danielkryska: See especially the README.md, have I described it clearly enough?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,9 @@ The most recent changes are on top, in each type of changes category.
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


jswk:
> shouldnt there be branches and PR in https://github.com/cyfronet-fid/docker-geoserver ? :)

Might be worth a try.

Issue #439, opened 2020-04-29
Create a loader of existing scenes from S3


jswk:

jswk:
The loader is for now placed in SeedProducts.
It is added as another dataset, which depends on s3.cloud.cyfronet.pl/s4e-sync-1 bucket contents.
So far, the schemas are loaded from resources and products base on them.
We should devise a workflow for production instances though, which should be rather based on admin accessible endpoints for managing Schemas, Products, and possibly Scenes.

jswk:
There is also a problem with showing Sentinel quicklooks in GeoServer - the responses are coming all transparent. Needs further investigation.

PR #447, opened 2020-05-06
[#439] Add s4e-sync-1 dataset

jswk:
The PR contains some refactoring prior to introducing the actual functionality.
The synchronization itself is a thin layer using PrefixScanner and SceneAcceptor and is not extracted to a separate class for now.

There is no endpoint for running this sort of synchronization. It should come with other endpoints for managing Schemas and Products, I guess.

#### New dataset implementation

Load the schemas from resources and create products which base on them in SeedProducts.
Then, PrefixScanner is used to enumerate all the keys under each product prefix.

The number of loaded scenes per product is controlled by an extra property: `seed.products.s4e-sync-1.limit`.

#### Correct S3 key handling and prevent connection leak

Keys must come without a leading slash, although we require the leading
slash in scene file to allow for relative path handling.

Execute a HeadObjectRequest in SceneStorage::exists, the opened
connection wasn't cleared and exhausted the AWS SDK connection pool.

#### Drop unique constraint on Scene.product_id and .timestamp

The constraint will not be met for Sentinel Scenes. However, it opens up the case for scene identification when updating, as these two columns are no longer unique.
A natural candidate seems to be the scene file key, I would hear some other opinions on this though.

Closes: #439.

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/db/seed/SchemaScanner.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/db/seed/SchemaScanner.java
@@ -1,0 +1,56 @@
+package pl.cyfronet.s4e.db.seed;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.ResourceLoader;
+import org.springframework.stereotype.Component;
+import pl.cyfronet.s4e.bean.Schema;
+import pl.cyfronet.s4e.util.ResourceReader;
+
+import javax.json.Json;
+import javax.json.JsonObject;
+import javax.json.JsonReader;
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+class SchemaScanner {
+    private final ResourceLoader resourceLoader;
+
+    public List<Schema> scan(String path) throws IOException {
+        log.trace(String.format("Scanning: '%s'", path));
+        Resource resource = resourceLoader.getResource(path);
+        File file = resource.getFile();
+        List<Schema> schemas = new ArrayList<>();
+        if (file.isDirectory()) {
+            for (String child : file.list()) {
+                schemas.addAll(scan(path + "/" + child));
+            }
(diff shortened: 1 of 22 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/db/seed/SchemaScanner.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+class SchemaScanner {
+    private final ResourceLoader resourceLoader;
+
+    public List<Schema> scan(String path) throws IOException {
+        log.info(String.format("Scanning: '%s'", path));
```
danielkryska, line 28: Is this log info necessary?

jswk: I will push it down to trace indeed.

danielkryska: ok

s4e-backend/src/main/java/pl/cyfronet/s4e/db/seed/SeedProducts.java
```diff
@@ -586,16 +684,41 @@ private void seedScenes(Product product, ProductParams params) {
         }
     }
 
-    private boolean objectExists(String key) {
-        HeadObjectRequest request = HeadObjectRequest.builder()
-                .bucket(s3Properties.getBucket())
-                .key(key)
-                .build();
+    private void readScenes(String prefix) {
+        // AWS SDK has a default connection pool size of 50, so 40 threads should be fine.
```
kmarszalek, line 688: <3

s4e-backend/src/test/java/pl/cyfronet/s4e/sync/PrefixScannerIntegrationTest.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.stream.Stream;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+
+/**
+ * Note that in these tests we pass it a path prefixed with slash.
+ * This is ok, most likely because of the minio implementation of the AWS S3, which is also different
+ * than CEPH.
+ * See e.g. https://github.com/minio/minio/issues/7717.
```
kmarszalek, line 18: :o

PR #451, opened 2020-05-11
[#450] Generic Tile View for admin and superadmin

### LINKED ISSUE #450: Generic Tile View for admin and superadmin
- [ ] Generic Tile View (see [admin](https://projects.invisionapp.com/share/S9RJNE8WXUB#/screens/414649284_Ustawienia) and [superadmin](https://projects.invisionapp.com/share/S9RJNE8WXUB#/screens/414671409_Ustawienia_-_Uberadmin) dashboards)
  - [ ] Superadmin Dashboard
  - [x] Admin Dashboard

danielkryska:
**IMPORTANT!!! For now tiles images/icons aren't available until @abacz update view.**

- Admin dashboard available on `zkAdmin@mail.pl` account (after going to `settings/profile` - `Panel główny` page)
https://projects.invisionapp.com/share/S9RJNE8WXUB#/screens/414649284_Ustawienia

- Super admin dashboard is available on `admin@mail.pl` (after going to `settings/profile` - `Panel główny` page)
https://projects.invisionapp.com/share/S9RJNE8WXUB#/screens/414671409_Ustawienia_-_Uberadmin

### DIFF
--- a/s4e-web/src/app/components/tiles-dashboard/tiles-dashboard.component.ts
+++ b/s4e-web/src/app/components/tiles-dashboard/tiles-dashboard.component.ts
@@ -1,0 +1,8 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 's4e-tiles-dashboard',
+  templateUrl: './tiles-dashboard.component.html',
+  styleUrls: ['./tiles-dashboard.component.scss']
+})
+export class TilesDashboardComponent {}
--- a/s4e-web/src/app/state/profile/profile.query.ts
+++ b/s4e-web/src/app/state/profile/profile.query.ts
@@ -1,15 +1,31 @@
-import {Injectable} from '@angular/core';
+import { Role } from './../session/session.model';
+import { Injectable } from '@angular/core';
 import {Query} from '@datorama/akita';
 import {ProfileStore} from './profile.store';
 import {ProfileState} from './profile.model';
 import {Observable} from 'rxjs';
 import {map} from 'rxjs/operators';
 
+const ROLE_INSTANCE_ADMIN = 'INST_ADMIN';
+const ROLE_INSTANCE_MANAGER = 'INST_MANAGER';
+const ROLE_GROUP_MANAGER = 'GROUP_MANAGER';
+
+const MANAGER_ROLES = [ROLE_INSTANCE_ADMIN, ROLE_INSTANCE_MANAGER, ROLE_GROUP_MANAGER];
+
+const hasAnyManagerRole = (roles: Role[]) => !!roles
+  .find(role => MANAGER_ROLES.includes(role.role));
+
 @Injectable({providedIn: 'root'})
 export class ProfileQuery extends Query<ProfileState> {
-
   constructor(protected store: ProfileStore) {
     super(store);
+  }
+
+  isAdmin() {
+    return this.getValue().admin;
+  }
+  isManager() {
+    return hasAnyManagerRole(this.getValue().roles);
   }
 
   public selectMemberZK(): Observable<boolean> {
@@ -17,7 +33,6 @@ export class ProfileQuery extends Query<ProfileState> {
   }
 
   public selectCanSeeInstitutions(): Observable<boolean> {
-    return this.select().pipe(map(state => state.admin ||
-      state.roles.find(role => ['INST_ADMIN', 'INST_MANAGER', 'GROUP_MANAGER'].includes(role.role)) != null));
+    return this.select().pipe(map(state => state.admin || hasAnyManagerRole(state.roles)));
   }
 }
--- a/s4e-web/src/app/utils.ts
+++ b/s4e-web/src/app/utils.ts
@@ -1,3 +1,5 @@
+import { InjectorModule } from './common/injector.module';
+import { ProfileQuery } from './state/profile/profile.query';
 import {UrlSegment} from '@angular/router';
 
 export function activateMatcher(url: UrlSegment[]) {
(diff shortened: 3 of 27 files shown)


s4e-web/src/app/components/tiles-dashboard/tiles-dashboard.component.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface ITileUrl {
+  label: string;
+  url: string;
+}
+
+export interface ITile {
+  header: string;
+
+  url?: ITileUrl;
+  urls?: ITileUrl[];
```
jswk, line 13: Why is there both url and url**s**?

s4e-web/src/app/utils.ts
```diff
@@ -12,3 +14,19 @@ export function activateMatcher(url: UrlSegment[]) {
...
+  return profileQuery.isAdmin && isDashboardUrl ? ({consumed: [url[0]]}) : null;
+}
+
+export function managerDashboardMatcher(url: UrlSegment[]) {
+  const profileQuery: ProfileQuery = AppInjector.get(ProfileQuery);
+  profileQuery.ngOnInit();
+  const lastUrlParameters = url[0];
+  const isDashboardUrl = !!lastUrlParameters && lastUrlParameters.path === 'dashboard';
+  return profileQuery.isManager && !profileQuery.isAdmin && isDashboardUrl ? ({consumed: [url[0]]}) : null;
+}
```
jswk, line 32: Are these needed? Couldn't router have overloaded paths?

s4e-web/src/app/views/settings/settings.routes.ts
```diff
@@ -23,7 +20,12 @@ export const settingsRoutes: Routes = [
...
       {
-        path: 'dashboard',
+        matcher: adminDashboardMatcher,
+        component: AdminDashboardComponent,
+        canActivate: [IsAdminGuard]
+      },
+      {
+        matcher: managerDashboardMatcher,
         component: DashboardComponent,
         canActivate: [IsManagerGuard]
```
jswk, line 30: I'm not sold on putting the admin/manager logic in between a matcher and a guard (see the other comment). If it's not possible to overload a path, I think the choice logic should be put in the routed component.

s4e-web/src/app/utils.ts
danielkryska: No, unfortunately it couldn't. 

1. Angular as a framework handle routing by going to the first path it achieves and then stops going through routes.
2. When routing is handled by Guard, it can be redirected to other sites, but will not continue to next (overloaded or not) path on rejection.
3. I'm not sure if in other versions than 7.x.x is it solved in another way, but in our one path and matcher can not be used together.

s4e-web/src/app/views/settings/settings.routes.ts
danielkryska: I was thinking about it a lot. In your last comment, I wrote why the matcher is used. Now I will try to describe, why guard and matcher are implemented together.

1. Matcher chooses a page where you can land.
2. Guard secure route against unauthorized access.

jswk: I disagree, matcher doesn't only choose the page, but also guards against someone unauthorized. I'd consider removing the guards, as they are redundant.

danielkryska: I've done now double-check. I don't see any comparison between both solutions (matcher and guard), for the first look it seems reasonable to use only one of them at once. Let's hope that the missing guard doesn't have any nuances that will propagate some bugs or lack of security in the future.

s4e-web/src/app-injector.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function setAppInjector(injector: Injector) {
+    if (AppInjector) {
+        // Should not happen
+        console.error('Programming error: AppInjector was already set');
+    }
+    else {
+        AppInjector = injector;
+    }
+}
```
michal-szostak, line 22: We already have the same mechanism in the `InjectorModul` you can remove this code, and use other one instead.

s4e-web/src/app/state/profile/profile.query.ts
```diff
@@ -7,11 +7,28 @@ import {map} from 'rxjs/operators';
...
 
+  public loadRole() {
+    this.isAdmin$().subscribe(isAdmin => this.isAdmin = isAdmin);
+    this.selectCanSeeInstitutions().subscribe(isManager => this.isManager = isManager);
+  }
+
+  public clearRole() {
+    this.isAdmin = false;
+    this.isManager = false;
+  }
```
michal-szostak, line 25: You can use store for this. And only provide methods like `isManager()` and `isAdmin()` instead of public fields.
Storing state in the query is not the best idea.

s4e-web/src/app/state/profile/profile.query.ts
```diff
@@ -7,11 +7,28 @@ import {map} from 'rxjs/operators';
...
+
+  isAdmin$(): Observable<boolean> {
```
michal-szostak, line 27: A minor one - in Akita Queries the convention for observable returning methods is `selectX...` so this would be something like `selectIsAdmin()`

s4e-web/src/app/state/session/session.query.ts
```diff
@@ -8,11 +8,20 @@ import {Observable} from 'rxjs';
...
   }
 
+  get isAdmin() {
+    return this._isAdmin;
+  }
+
+  set isAdmin(isAdmin: boolean) {
+    this._isAdmin = isAdmin;
+  }
+
```
michal-szostak, line 24: The same case as the one in `ProfileQuery` this can be stored in store

s4e-web/src/app/views/settings/dashboard/dashboard.component.ts
```diff
@@ -1,15 +1,15 @@
...
-export class DashboardComponent implements OnInit {
+export class DashboardComponent {
+  constructor(private _profileQuery: ProfileQuery) {}
 
-  constructor() { }
-
-  ngOnInit() {
+  get isAdmin() {
+    return this._profileQuery.isAdmin$();
   }
```
michal-szostak, line 14: calling method which returns observable in a getter of the component is a very bad idea, which may lead to a lot of hard to detect bugs. Is you need `isAdmin` observable in the component you should assign it in the `ngOnInit`

s4e-web/src/app/state/session/session.query.ts
```diff
@@ -8,11 +8,20 @@ import {Observable} from 'rxjs';
...
     super(store);
   }
 
+  get isAdmin() {
+    return this._isAdmin;
+  }
+
+  set isAdmin(isAdmin: boolean) {
+    this._isAdmin = isAdmin;
+  }
```
michal-szostak, line 23: This should be taken from store

Issue #459, opened 2020-05-13
UI/API version of application


michal-szostak:
The version of the application should be read from optional environmental variable.

The variable passing the version number should be `NG_VERSION`.

When building application the version should be passed as follows:

```
NG_VERSION='1.0.0' npm run build
```

Or

```
NG_VERSION='1.0.0' ./mvnw package
```

PR #460, opened 2020-05-13
[#459] Application version

michal-szostak:
## What is in this PR?

* Add baking of application version (provided by env variable during compile time)

## How does it work? 

The mechanism is described in the issue description #459, but the TLDR is that you should build / run application while providing version with environmental variable `NG_VERSION`. This env variable can be supplied while running `npm` or `mvn`.

## How does it look like?

For simplicity the version is outputted in the console when application starts:

image

If we need version to be accessible only to developers / testers this should be good enough, otherwise there should be some place in the application where version is being displayed, I'll leave it to @abacz , it can be added later as part of some small issue. 

## Other considerations

It is very probable that the initial PR will fail build, as I made it necessary to have specified version when building production. @wziajka or @jswk - it would be great if `mvnw` package could provide this value, otherwise I would ask @wziajka to modify `.travis.yml` to properly set the variable (in the same way you'll do this for our staging build.

I can also disable this mechanism, and always have some kind of default, waiting for opinions

Closes: #459


jswk:
I think this variable should be set in `s4e-web/pom.xml`. Currently, default version is 0.0.0-SNAPSHOT, and it is composed of two parts ${revision}${changelog}. There is a third part, which can be used for this: ${sha1}, then the version could be composed as ${revision}${sha1}${changelog}, e.g. 0.0.0-123908de7865ae-SNAPSHOT.
The middle part would be supplied from `GITHUB_SHA` envvar, see. https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables.
And then, the complete version would be passed to `NG_VERSION` envvar in `npm run-script build` front-end plugin execution.

I would do this in an emergent issue. Just please modify this one not to fail on missing `NG_VERSION`.

s4e-web/src/app/components/root/root.component.ts
```diff
@@ -13,6 +13,7 @@ export class RootComponent {
   PRODUCTION: boolean = environment.production;
 
   constructor(private modalService: ModalService, private notificationService: NotificationService) {
+    console.log(`Sat4Env version ${environment.version}`);
```
jswk, line 16: Typo

jswk:
Okay, #461 is merged to master. You can rebase and verify it's working fine together.

s4e-web/src/app/components/root/root.component.ts
```diff
@@ -13,6 +13,7 @@ export class RootComponent {
   PRODUCTION: boolean = environment.production;
 
   constructor(private modalService: ModalService, private notificationService: NotificationService) {
+    console.log('Sat4Env version ' + environment.version);
```
jswk, line 16: There's still a typo here.

PR #465, opened 2020-05-14
[#440] Listen to scene files notifications from AMQP queue

jswk:
I'll update the description on Monday.

Fixes: #440.


README.md
```diff
@@ -187,6 +187,18 @@ To control the number of scenes seeded per product for `s4e-sync-1` dataset use
 To load an unlimited number set its value to -1.
 
 
+#### Scene updates from AMQP queue
+
+Application can be configured to listen to notifications from an AMQP queue with S3 object updates.
+
+In profile `development` this feature is enabled, and in `production` it has to be enabled by setting
+`amqp.enabled=true` or `AMQP_ENABLED=true`.
+
```
kmarszalek, line 196: i would add that you can configure queues names by properties.

s4e-backend/src/main/java/pl/cyfronet/s4e/sync/QueueReceiver.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.amqp.rabbit.annotation.RabbitListener;
+
+@RequiredArgsConstructor
+@Slf4j
+public class QueueReceiver {
+    private final SceneAcceptor sceneAcceptor;
+
+    @RabbitListener(queues = "#{incomingQueueName}")
+    public void receive(Notification notification) {
+        log.info(String.format("Handling %s: %s", notification.getEventName(), notification.getObjectKey()));
```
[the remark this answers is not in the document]
jswk: I'd argue that we should log that we have received a message from queue. It will by useful to identify problems early on, however, it should probably be done in DEBUG level, you're right.
I'll also correct the message.

README.md
jswk: Will do

PR #466, opened 2020-05-14
454 new sidebar html css

abacz:


s4e-web/src/assets/images/ico_drag.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
danielkryska, line 1: Have you maybe think about using [Font Awesome Icons](https://fontawesome.com/icons) and tag <i></i> in your code? It's much easier to put into an application than loading SVG files into assets and loading it trough image src parameter.

abacz: Yes i think about FA but:
1. I try to avoid inserting additional tag into html where is not necessary.   
2. Not always using <i> is easier than adding svg to css.

Besides I don't want to be forced to use it in my design as an only solution. ;)
Part of icons are made by myself where i don't have same "look and feel" design in FA.

Of course we can mixed solutions.

kmarszalek:
do you want to merge this without favorites?

jswk:
Another thing I noticed, but it is more @michal-szostak and @danielkryska thing, is the working of the search input.
Try this:
1. Type Krak into the search. You'll see results:
image
2. Click on Kraków. The map will center on Kraków and the input will read `Kraków`.
image
3. Click on the looking glass. The map view will show Krakowiany instead.
image

jswk:
There is a problem with small window height though.
image
The bottom products are hidden behind the layers region and there is no scroll. The scroll actually appears if the layers list is taller than the browser height, but the bottom products remain inaccessible.

s4e-web/src/assets/images/ico_drag.svg
danielkryska: ok

danielkryska:
@jswk my newest PR is disabling `loupe` when results are not opened, so users will not be misled.

Issue #477, opened 2020-05-21
Refactor aside.js from sidebar


michal-szostak:
This is a follow up issue of #454 (@abacz inserted aside.js to illustrate functionality and do UX testing, the code is unobstrusive and can be easily removed).

This issue requires two things:
- [x] Refactor `src/assets/js/aside.js` into the sidebar code (possibly as a directive)
- [x] Remove reference to `aside.js` from `index.html`

PR #488, opened 2020-06-02
[#477] Refactor aside.js from sidebar

danielkryska:
* remove aside.js
* sidenavbar on map should have scroll on height resize (when layers overlap products)

Closes: #477


danielkryska:
2. That's the issue for @abacz #489 
3. Fixed
4. Fixed

Issue #479, opened 2020-05-21
GeoServer licensing gateway


jswk:
Create a GeoServer licensing gateway, which will be put in front of GeoServer instances. Each WMS request should go through it and be authorized based on JWT.

Gateway should be configurable from properties, from my analysis it follows, that DB access is non-essential.
Apart from connection specification, it will need a list of open and EUMETSAT licensed layers (identified by WMS layer names).
Excerpt from a proposed configuration:
```properties
layers.setvak_pl=open
layers.setvak_eu=open
layers.108m=EUMETSAT
```

Each request would be filtered based on the following logic.
```python
if req.layer is open:
    allow
if req.layer is EUMETSAT && req.time is not fresh:
    allow
if req.layer in jwt.layers:
    allow
else:
    reject
```
In this way, JWT is only unwrapped if necessary, although it's signature should've been validated by HAProxy, so that part may be omitted.
`req.time` is considered fresh if `req.time > now() - 3h`.

The licensing info has to be kept in sync with backend, however, is shouldn't change that often. By default products are assumed to be private, unless marked open or EUMETSAT. Backend manages access to private products by appropriate JWT generation.

(The backend part should be considered in #288. Also #108 has some prior discussion.)

jswk:
Finally, the configuration was changed to:
```
gateway.open-layers=development:setvak,development:wv_ir
gateway.eumetsat-layers=development:108m
```

The key based method is fine when using .properties files, but problematic with envvar configuration (in case of docker-compose).

PR #490, opened 2020-06-03
GeoServer licensing gateway

jswk:
Add gs-gateway module, which contains a gateway which authorizes requests to GeoServer based on JWT token.

The token will contain a `layers` claim, which lists layers to which access should be granted regardless of other policies.

The gateway is configured with lists of `OPEN` and `EUMETSAT` layers.
`OPEN` means requests to it are passed without exception.
`EUMETSAT` means that requests with TIME param less than 3h old are discarded.
Requests to all other non-listed layers are discarded, save the JWT claim `layers` contains it.

Other than that, all params other than specified in the config are stripped from the requests, so no `CQL_FILTER`, for example, will go through.

This module is wired up before GeoServer instance, so all requests coming from the clients must go through it. Requests from backend reach GeoServer directly though.

## Additions

Other than the main module, I refactored the maven modules a little. Root module no longer inherits from spring-boot-parent, so things had to be configured a little differently in some places. Also, to override a dependency version specified in spring-boot BOM, you will now have to override it in dependencyManagement, so that it overrides BOM.

@wziajka: Have a look at the docker and docker-compose mostly. Any thoughts on this way of configuring the new module?

Closes: #479.


kmarszalek:
looked today and yesterday at this, it looks fine, but before giving review i need to sleep on it one more time :)

jswk:
> where is changelog xD ?

Haha, missed it. Now it's in place.

PR #495, opened 2020-06-05
494 edit institution

danielkryska:
- [x] Edit institution in `settings/institution`
    - [x] reset store errors before each request
    - [x] clear form after submit and navigate to `/institutions-management` or `/institution-profile`
    - [x] Institution profile
    - [x] load and convert image from `static/emblems` to base64


s4e-web/src/app/views/settings/state/institution/institution.utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    .some(institution => institution.slug.indexOf(activeInstitutionSlug) > -1);
+  if (!institutionExists) {
+    store.setError('no_institution');
+    return null;
+  }
+
+  store.setActive(activeInstitutionSlug);
+  store.setError(null);
+  return activeInstitutionSlug;
+}
```
michal-szostak, line 58: I'm strongly against moving this out of the service class. I can understand will to make classes less bloated, but this does not solve any problem and only moves code around and makes it harder to find.

Also - are those functions working correctly? I see calls to `this.store` and other `this.*`

Issue #508, opened 2020-06-10
Change add group and add person to modal


danielkryska:
- [ ] Change add group and add person to modal
  - [ ] Refactor add person
  - [ ] Write tests for add person
  - [ ] Change add person to modal
  - [x] Change add group to modal
  - [x] Remove Group on `trash icon` click

PR #514, opened 2020-06-12
[#508] Change add group and add person to modal

danielkryska:
- [x] Change add group and add person to modal
  - [x] Refactor add person
  - [x] Write tests for add person
  - [x] Change add person to modal
  - [x] Change add group to modal
  - [x] Remove Group on `trash icon` click

### DIFF
--- a/s4e-web/src/app/views/settings/people/person-form/person-form.component.html
+++ b/s4e-web/src/app/views/settings/people/person-form/person-form.component.html
@@ -1,34 +1,37 @@
-<header class="page__header">
-  <h1 i18n>Dodaj nową osobę</h1>
-</header>
 
-<section class="page__content">
-  <form [formGroup]="form" class="page__form">
-    <ul>
-      <li>
-        <label for="name" i18n>Imię</label>
-        <input id="name" type=text formControlName="name">
-      </li>
-      <li>
-        <label for="surname" i18n>Nazwisko</label>
-        <input id="surname" type=text formControlName="surname">
-      </li>
-      <li>
-        <label for="email" i18n>Email</label>
-        <input id="email" type="email" formControlName="email">
-      </li>
-      <li>
-        <strong i18n>Grupy</strong>
-        <s4e-check-box-list [items]="groups$ | async" formControlName="groupSlugs"></s4e-check-box-list>
-        <ng-template #groupsLoading>
-          Ładowanie...
-        </ng-template>
-      </li>
-      <li class="actions">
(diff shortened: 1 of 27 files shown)


s4e-web/src/app/views/settings/people/person-form/person-form.component.html
```diff
@@ -1,34 +1,37 @@
...
+      <form [formGroup]="form" class="page__form">
+        <ul>
+          <li>
+            <ext-input formControlName="name" label="Imię" i18n></ext-input>
+          </li>
+          <li>
+            <ext-input formControlName="surname" label="Nazwisko" i18n></ext-input>
+          </li>
+          <li>
+            <ext-input type="email" formControlName="email" label="Email" i18n></ext-input>
```
jswk, line 16: If `!!person` then these inputs should be disabled, as you'd be editing an existing user.
Also, I noticed when I edit these that `POST /institutions/{institution}/users` request is sent. This should actually result in some error on the backend side, which it doesn't (that's an issue on backend side, I added its description [here](https://github.com/cyfronet-fid/sat4envi/issues/517#issuecomment-644236048)), but front should instead send `PUT /institutions/{institution}/users`

danielkryska: I've updated fields to be disabled in edition mode and POST has been changed to PUT.
In case of an invalid payload that is sending with the form to API while edition, I'll just leave it until we'll have any conception of people and group management.

jswk:
Something changed since I've accepted this?

danielkryska:
> Something changed since I've accepted this?

Few lines in input component

jswk:
> > Something changed since I've accepted this?
> 
> Few lines in input component

How come since nothing was pushed in between my accepting and you re-requesting?

danielkryska:
> > > Something changed since I've accepted this?
> > 
> > 
> > Few lines in input component
> 
> How come since nothing was pushed in between my accepting and you re-requesting?

My mistake, I've two tickets opened while requesting re-review from you. Nothing changed since...

s4e-web/src/app/views/settings/people/person-form/person-form.component.spec.ts
```diff
@@ -17,19 +28,48 @@ describe('PersonFormComponent', () => {
...
   it('should create', () => {
+    fixture.detectChanges();
     expect(component).toBeTruthy();
   });
+  it('should add person', fakeAsync(() => {
+    const person = {
+      email: 'test@mail.pl',
+      name: 'test',
+      surname: 'test'
+    };
```
michal-szostak, line 56: This should be a factory

PR #518, opened 2020-06-14
493 instgroup improvement html css

abacz:


abacz:
Not ready. Please do not review.

s4e-web/src/app/form/check-box-list/check-box-list.component.scss
```diff
@@ -1 +1 @@
-@import "../../../scss/panel";
+@import "../../../scss/settings";
```
danielkryska, line 1: I see a lot of such imports. Most of them are the only line of a component's CSS. I suppose you are using it in such way to encapsulate their styles. Have you maybe wonder if it would be easier for you to start working on global CSS without importing them into specific files?

danielkryska: Never mind, @michal-szostak said that you talked already about this issue... :P

kmarszalek:
- button "dodaj jednostkę podrzędną" doesn't work.
- changelog
- new institution -> anuluj -> what should happen ? now it redirects to root institution.
nice squash :)

abacz:
- "dodaj jednostkę" is static for further development
- i changed changelog
- "anuluj" in new institution should redirect to institution list

PR #525, opened 2020-06-17
[#503] Create dynamic breadcrumbs and add general breadcrumb in /settings

### LINKED ISSUE #503: Dynamic breadcrumbs + add general breadcrumb in /settings/*
- [ ] Create a dynamic breadcrumb component loading breadcrumbs from `activated route`
- [ ] Add breadcrumbs to `route data` param
- [ ] Add general breadcrumb to /settings/*

danielkryska:
[#503] Create dynamic breadcrumbs and add general breadcrumb in /settings

* Create a dynamic breadcrumb component loading breadcrumbs from activated route
* Add breadcrumbs to route data param
* Add general breadcrumb to /settings/*

Closes #503
Screenshot_2020-06-19 Sat4Envi - Dane satelitarne na wyciągnięcie ręki(1)
Screenshot_2020-06-19 Sat4Envi - Dane satelitarne na wyciągnięcie ręki

Screenshot_2020-06-19 Sat4Envi - Dane satelitarne na wyciągnięcie ręki(2)

### DIFF
--- a/s4e-web/src/app/views/settings/breadcrumb/breadcrumb.component.html
+++ b/s4e-web/src/app/views/settings/breadcrumb/breadcrumb.component.html
@@ -1,0 +1,15 @@
+<div class="breadcrumb">
+  <span i18n>Tutaj jesteś: </span>
+  <span>
+    <ng-container *ngFor="let breadcrumb of breadcrumbs; last as isLast">
+      <ng-container *ngIf="isLast; else parent">
+        {{ breadcrumb.label }}
+      </ng-container>
+      <ng-template #parent>
+        <a [routerLink]="breadcrumb.url" queryParamsHandling="preserve">{{ breadcrumb.label }}</a> >
+      </ng-template>
+    </ng-container>
+  </span>
+</div>
+
+
--- a/s4e-web/src/app/views/settings/breadcrumb/breadcrumb.component.ts
+++ b/s4e-web/src/app/views/settings/breadcrumb/breadcrumb.component.ts
@@ -1,0 +1,37 @@
+import { untilDestroyed } from 'ngx-take-until-destroy';
+import { IBreadcrumb } from './breadcrumb.model';
+import { filter } from 'rxjs/operators';
+import { BreadcrumbService } from './breadcrumb.service';
+import { ActivatedRoute, Router, NavigationEnd } from '@angular/router';
+import { Component, OnInit, OnDestroy } from '@angular/core';
+
+@Component({
+  selector: 's4e-breadcrumb',
+  templateUrl: './breadcrumb.component.html',
+  styleUrls: ['./breadcrumb.component.scss']
+})
+export class BreadcrumbComponent implements OnInit, OnDestroy {
+  public breadcrumbs: IBreadcrumb[];
+
+  constructor(
+    private _router: Router,
+    private _activatedRoute: ActivatedRoute,
+    private _breadcrumbService: BreadcrumbService
+  ) {}
+
+  ngOnInit() {
+    this._breadcrumbService.loadFrom(this._activatedRoute);
+
+    this._router.events
+      .pipe(
+        untilDestroyed(this),
+        filter(event => event instanceof NavigationEnd)
+      )
+      .subscribe(() => this._breadcrumbService.loadFrom(this._activatedRoute));
+    this._breadcrumbService.breadcrumbs$
+      .pipe(untilDestroyed(this))
+      .subscribe((breadcrumbs) => this.breadcrumbs = breadcrumbs);
+  }
+
+  ngOnDestroy() {}
+}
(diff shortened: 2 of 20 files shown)


s4e-web/src/app/views/settings/breadcrumb/breadcrumb.component.html
```diff
@@ -0,0 +1,17 @@
...
+    <ng-container *ngFor="let breadcrumb of breadcrumbs$ | async; index as index$">
+      <ng-container [ngSwitch]="index$">
+        <ng-container *ngSwitchCase="(breadcrumbs$ | async).length - 1">
+          {{ breadcrumb.label }}
+        </ng-container>
+        <ng-container *ngSwitchDefault>
+          <a [routerLink]="breadcrumb.url" queryParamsHandling="preserve">{{ breadcrumb.label }}</a> >
+        </ng-container>
+      </ng-container>
+    </ng-container>
```
michal-szostak, line 13: You can do it in a much simpler way by using `first` and `last` (these are available in the same way as `index`

```
<ng-container *ngFor="let breadcrumb of breadcrumbs$ | async; last as last$">
```

`first` / `last` are normal `boolean` variables

s4e-web/src/app/views/settings/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  constructor(
+    private _router: Router,
+    private _activatedRoute: ActivatedRoute,
+    private _breadcrumbService: BreadcrumbService
+  ) {}
+
+  ngOnInit() {
+    this._router.events
+      .pipe(filter(event => event instanceof NavigationEnd))
```
michal-szostak, line 24: `untilDestoryed` should go somewhere here

And also - you have nested subscription here so some other mechanism to unsubscribe may be required

s4e-web/src/app/views/settings/breadcrumb/breadcrumb.service.ts
```diff
@@ -0,0 +1,105 @@
...
+  registerRoutes(childrenRoutes: Routes, rootPath = '') {
+    this.defaultRoute = this._getDefaultRouteFrom(childrenRoutes);
+    this.rootPath = rootPath;
+  }
+  loadFrom(activatedRoute: ActivatedRoute) {
+    return this._getBreadcrumbsFrom(this._getActualRouteFrom(activatedRoute))
+      .subscribe((breadcrumbs: IBreadcrumb[]) => this._breadcrumbSubject$.next(breadcrumbs));
+  }
+
+  _getBreadcrumbsFrom(childRoute: ActivatedRoute, breadcrumbs: IBreadcrumb[] = [], depth = 0): Observable<IBreadcrumb[]> {
```
michal-szostak, line 32: If this is private then mark it as `private` (or `protected`), same for others

s4e-web/src/app/views/settings/breadcrumb/breadcrumb.service.ts
```diff
@@ -0,0 +1,105 @@
...
+  _getDefaultRouteFrom(routes: Routes): Route {
+    const redirectRoute = routes.find((route: Route) => route.path === '**');
+    return !!redirectRoute
+      ? routes.find((route: Route) => route.path === redirectRoute.redirectTo)
+      : null;
+  }
+  _withDefault$(breadcrumbs: IBreadcrumb[]): Observable<IBreadcrumb[]> {
+    if (!this._isStartPageIn(breadcrumbs)) {
+      this._addNew(this.defaultRoute, breadcrumbs);
+      return of(breadcrumbs);
```
michal-szostak, line 62: you return `of(breadcrumbs` regardless, so I think this line is redundant

s4e-web/src/app/views/settings/settings.routes.ts
```diff
@@ -1,88 +1,166 @@
...
-        path: 'institutions',
-        component: InstitutionListComponent,
-        canActivate: [IsManagerGuard]
+        canActivate: [IsManagerGuard],
+        data: {
+          breadcrumbs: [
+            {
+              label: 'Grupy'
+            }
+          ] as IBreadcrumb[]
```
michal-szostak, line 124: I'm not 100% sure it will work (the AOT compiler can be somewhat tricky with these things)

But I would suggest making a helper function - something like this:

```
export function makeBreadcrumbs(...breadcrumbs: (IBreadcrumb|string)[]): IBreadcrumb[] {
  return breadcrumbs.map(breadcrumb => typeof breadcrumb === 'string' ? {label: breadcrumb} : breadcrumb);
}
```

Then you can define breadcrumbs simply by:

```
breadcrumbs: makeBreadcrumbs('Grupy')
```

s4e-web/src/app/views/settings/breadcrumb/breadcrumb.component.html
danielkryska: It isn't much simpler, but it is good to know it :)

s4e-web/src/app/views/settings/breadcrumb/breadcrumb.service.ts
danielkryska: The problem occurs while testing. I can't access public methods and testing the main function, because it is too complex and recursive. On other hand, I want to say other developers that those methods are created only for internal usage. If you have any proposal in such case, I'm open on it...

AndrzejWojcik:
Manual tests have passed.

PR #567, opened 2020-07-06
Backend dependencies update - 2020.07

jswk:
Disable default swagger url.

Move jsonb hibernate type declaration to interested classes, as keeping
it in package-info.java started causing static initialization issues
when starting an application from IntelliJ configuration.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,20 +26,20 @@
         <java.version>11</java.version>
         <maven.compiler.release>11</maven.compiler.release>
 
-        <spring-boot.version>2.3.0.RELEASE</spring-boot.version>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -26,20 +26,20 @@
         <java.version>11</java.version>
         <maven.compiler.release>11</maven.compiler.release>
 
-        <spring-boot.version>2.3.0.RELEASE</spring-boot.version>
+        <spring-boot.version>2.3.1.RELEASE</spring-boot.version>
 
-        <awssdk.version>2.13.31</awssdk.version>
+        <awssdk.version>2.13.50</awssdk.version>
```
kmarszalek, line 31: 51?

Issue #583, opened 2020-07-08
Cookie based authorization


jswk:
Apart from reading the token from `Authorization` header, check cookie `token` for it (both in s4e-backend and gs-gateway).

The cookie should be set in `/login` endpoint and reset in `/logout`.
The preexisting `/login` endpoint will be moved to `/token`.

PR #587, opened 2020-07-10
[#583] Cookie based auth front end

michal-szostak:
## What is in this PR?

* Change authentication schema to cookies
* Fix circular dependencies
* Merge Session and Profiles features into one

Closes: #583


s4e-web/src/app/state/session/session.service.ts
```diff
@@ -1,97 +1,102 @@
...
   @action('login')
   login(formState: LoginFormState) {
-    this.sessionStore.setLoading(true);
+    this.store.setLoading(true);
     const {login: email, password, ...x} = formState;
-    const url = `${this.CONFIG.apiPrefixV1}/token`;
-    this.http.post<LoginRequestResponse>(url, {email, password},
-      {headers: {[ERROR_INTERCEPTOR_CODES_TO_SKIP]: `${HTTP_404_BAD_REQUEST},${HTTP_401_UNAUTHORIZED}`} })
+    const url = `${this.CONFIG.apiPrefixV1}/login`;
+    this.http.post(url, {email, password},
```
danielkryska, line 69: Why are you not using httpPostRequest$ functionality with 
```
.pipe(        
  switchMap(data => this.getProfile$()),
  tap(() => this.store.update({email: formState.login}))
)
```

s4e-web/src/app/state/session/session.service.ts
```diff
@@ -1,97 +1,102 @@
...
+
+  clearError() {
+    this.store.setError(null);
   }
 
   private _navigateToApplication(): void {
+    const router: Router = InjectorModule.Injector.get(Router);
+
     !!this._backLink && this._backLink !== ''
-      ? this.router.navigateByUrl(this._backLink)
```
danielkryska, line 93: Why haven't you use context `this.router`?

s4e-web/src/app/utils/initializer/config.service.spec.ts
```diff
@@ -1,44 +1,49 @@
...
-      done();
-    });
-
-    const req = http.expectOne('api/v1/config');
+  it('should call endpoint', async () => {
+    const spy = spyOn(sessionService, 'getProfile$').and.returnValue(of(SessionFactory.build()));
+    const p = configService.loadConfiguration();
 
-    req.flush({
+    const reqConfig = http.expectOne('api/v1/config');
```
danielkryska, line 32: Why you're not using `s4eConfig` service to get `API_PREFIX`?

s4e-web/src/app/utils/initializer/config.service.ts
```diff
@@ -35,14 +33,16 @@ export class S4eConfig implements IConfiguration {
...
-      this.profileService.get$()
-    ]).pipe(
-      map(([configuration, profile]) => configuration)
-    ).toPromise();
+  loadConfiguration(): Promise<S4eConfig> {
+    return this.http.get<IRemoteConfiguration>(`api/v1/config`)
+      .pipe(
+        tap(config => this.init(config)),
+        switchMap(() => this.sessionService.getProfile$()),
+        map(() => this)
```
danielkryska, line 45: What are you doing in this line: `map(() => this)`?

s4e-web/src/app/views/map-view/map/map.component.ts
```diff
@@ -26,37 +24,37 @@ import ImageWrapper from 'ol/Image';
...
     private _loaderService: NgxUiLoaderService,
     private _notificationService: NotificationService,
     private _sessionQuery: SessionQuery
-  ) {}
+  ) {
+  }
+
+  private _overlays: UIOverlay[] = [];
+
+  get overlays(): UIOverlay[] {
```
danielkryska, line 50: I've heard a lot of bad opinions about using `getters` and `setters` in typescript (better option would be `getOverlays()` function). They can provide hard to find bugs and inconsistent behaviors. Is there any reason why you're using it?

s4e-web/src/app/views/map-view/map/map.component.ts
```diff
@@ -118,6 +116,41 @@ export class MapComponent implements OnInit, OnDestroy {
     });
   };
 
+  public getMapData(): Observable<MapData> {
+    this.working.emit(true);
+    const canvas: HTMLCanvasElement = this.map.getViewport().firstChild as HTMLCanvasElement;
+
+    const r = new ReplaySubject<MapData>(1);
```
danielkryska, line 123: One time `r` is `request`, another `r` is `mapData$` in case of consistency and readability could you change name of this variable?

s4e-web/src/app/views/map-view/map/map.component.ts
```diff
@@ -118,6 +116,41 @@ export class MapComponent implements OnInit, OnDestroy {
     });
   };
 
+  public getMapData(): Observable<MapData> {
+    this.working.emit(true);
+    const canvas: HTMLCanvasElement = this.map.getViewport().firstChild as HTMLCanvasElement;
```
danielkryska, line 121: Could you change `canvas` to `mapViewportCanvas` or `mapCanvas`?

s4e-web/src/app/views/map-view/map/map.component.ts
```diff
@@ -118,6 +116,41 @@ export class MapComponent implements OnInit, OnDestroy {
...
+
+    const r = new ReplaySubject<MapData>(1);
+
+    // this will work only on new browsers!
```
danielkryska, line 125: I see the third time a needed comment in my whole life. Keep it up!!! :) 

Could you add prefix like `IMPORTANT!!!` to focus other developers attention?

s4e-web/src/app/views/map-view/map/map.component.ts
```diff
@@ -26,37 +24,37 @@ import ImageWrapper from 'ol/Image';
...
-  private _overlays: UIOverlay[] = [];
-
-  private overlays$: BehaviorSubject<UIOverlay[]> = new BehaviorSubject([]);
   @Output() viewChanged = new EventEmitter<ViewPosition>();
-  private baseLayer: Layer;
-  private map: Map;
-  private activeScene$ = new ReplaySubject<Scene | null>(1);
   @ViewChild('linkDownload', {read: ElementRef}) linkDownload: ElementRef;
   @Input() isWorking: boolean = false;
   @Output() working = new EventEmitter<boolean>();
```
danielkryska, line 30: Instead of having two different outputs and inputs, of `isWorking`, could you merge them as one (, because of their one functionality)? 

```
Example
@Output('isWorkingChange') emitter: EventEmitter<boolean> = new EventEmitter<boolean>();
@Input('isWorking') set isWorking(value) { ... }

```

s4e-web/src/app/state/session/session.service.ts
michal-szostak: Because it dos not support custom headers which are being used. It will have to be taken into the consideration when doing general refactoring of http functions

s4e-web/src/app/state/session/session.service.ts
michal-szostak: Because of circular dependency. This is a small hack, which probably should be refactored as part of separate PR. (separate `S4EConfig` from config loader should break circular dependency

s4e-web/src/app/utils/initializer/config.service.spec.ts
michal-szostak: That's one place where it is hardcoded, as we're testing it here. At this point It is not initialized.

s4e-web/src/app/utils/initializer/config.service.ts
michal-szostak: `loadConfiguration()` is returning `S4EConfig` instance so I need to return it. I'm doing switch map earlier to load Profile information, so I need to change the context.

s4e-web/src/app/views/map-view/map/map.component.ts
michal-szostak: I do agree, but I think that it's outside of the scope of this issue. The reason is that those inputs / outputs were added at the different times.

But I'll write it down for future improvements.

Issue #601, opened 2020-07-15
'Brak Wyników' is visible in sentinel search results even when results are present


michal-szostak:
![image](https://user-images.githubusercontent.com/13235269/87538660-d93be380-c69c-11ea-8b60-7d5947790e7d.png)

PR #602, opened 2020-07-15
[#601] No Search Result is showing even when results are found

michal-szostak:
## What is in this PR?

Fix bug with 'Brak wyników' being show regardless of whether sentinel search results were present.
Added unit test coverage to prevent regression.

Closes: #601

### DIFF
--- a/s4e-web/src/app/test.utils.spec.ts
+++ b/s4e-web/src/app/test.utils.spec.ts
@@ -1,0 +1,11 @@
+import {Observable} from 'rxjs';
+import {take, toArray} from 'rxjs/operators';
+
+export function toTestPromise<T>(observable: Observable<T>, takeLast: number = 1): Promise<T|T[]> {
+  observable = observable.pipe(take(takeLast))
+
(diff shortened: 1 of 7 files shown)


s4e-web/src/app/test.utils.spec.ts
```diff
@@ -0,0 +1,11 @@
+import {Observable} from 'rxjs';
+import {take, toArray} from 'rxjs/operators';
+
+export function toTestPromise<T>(observable: Observable<T>, takeLast: number = 1): Promise<T|T[]> {
+  observable = observable.pipe(take(takeLast))
+
+  if (takeLast === 1) {
+    return observable.toPromise();
+  }
+  return observable.pipe(toArray()).toPromise();
```
danielkryska, line 10: This can be simplified 

```
 if (takeLast === 1) {
    return observable.toPromise();
  }
  return observable.pipe(toArray()).toPromise();
```

to

```
  return (takeLast === 1 ? observable : observable.pipe(toArray()) ).toPromise();
```

PR #607, opened 2020-07-16
[#512] Query param handling in search endpoint

### LINKED ISSUE #512: Query param handling in search endpoint
Some cases that should be handled as they now result in 500:
- `limit=-1`
- `limit=abc`
- `offset=-1`
- `offset=abc`

Apart from these basics, format of parameters should be verified: date-time, cloudCover (this should also be limited to range [0,100]).

All such errors should be handled in a unified way, possibly similar to how Spring handles `@Valid`. In this case we're unable to utilize the spring built-in feature, but the builiding blocks remain available.

Create method analogous to `pl.cyfronet.s4e.ex.ErrorHandlerHelper#toResponseMap(org.springframework.web.bind.MethodArgumentNotValidException)`, but it should utilize our type of exception, which would also contain a BindingResult with validation problem with params (i.e. limit, offset, and others).

Validation should be done by calls to Errors object, MapBindingResult should be fine to carry validation information - it should be passed to the QueryBuilder method from PreparedStatementBuilder. PSB would in turn throw appropriate exception encapsulating this BindingResult if there were some errors encountered when building the query, If, on the other hand exception occurs when executing it, probably only the `__general__` should be populated, also - it might actually make sense to return 500 in such case because that would mean we didn't validate something correctly or the DB is somehow corrupted.

The validation should take place in the QueryBuilders, which know the context the best. Also, this will allow for easy testing of the validation.

The exception thrown in PSB would propagate to ErrorHandler, which would turn it into a proper error message.

kmarszalek:
Closes #512.

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/data/repository/query/QueryNumber.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/data/repository/query/QueryNumber.java
@@ -1,11 +1,16 @@
 package pl.cyfronet.s4e.data.repository.query;
 
+import org.springframework.validation.Errors;
+
 import java.util.List;
 import java.util.Map;
 
 import static pl.cyfronet.s4e.search.SearchQueryParams.CLOUD_COVER;
 
 public class QueryNumber extends QueryDecorator {
+    public static final int MAX_CLOUD_COVERAGE = 100;
+    public static final int MIN_CLOUD_COVERAGE = 0;
+
     public QueryNumber(QueryBuilder queryBuilder) {
         super(queryBuilder);
     }
@@ -13,16 +18,32 @@ public class QueryNumber extends QueryDecorator {
     @Override
     public void doPrepareQueryAndParameters(Map<String, Object> params,
                                             List<Object> parameters,
-                                            StringBuilder resultQuery) {
-        queryByCloudCover(params, parameters, resultQuery);
+                                            StringBuilder resultQuery,
+                                            Errors errors) {
+        queryByCloudCover(params, parameters, resultQuery, errors);
     }
 
     private void queryByCloudCover(Map<String, Object> params,
                                    List<Object> parameters,
-                                   StringBuilder resultQuery) {
+                                   StringBuilder resultQuery,
+                                   Errors errors) {
         if (params.containsKey(CLOUD_COVER)) {
             resultQuery.append(" AND (metadata_content ->> 'cloud_cover')::float <= ? ");
-            parameters.add(Float.parseFloat(params.get(CLOUD_COVER).toString()));
+            parameters.add(getCloudCover(params, errors));
+        }
+    }
+
+    private float getCloudCover(Map<String, Object> params,
+                                Errors errors) {
+        try {
+            float result = Float.parseFloat(params.get(CLOUD_COVER).toString());
+            if (result < MIN_CLOUD_COVERAGE || result > MAX_CLOUD_COVERAGE) {
+                throw new NumberFormatException("Cloud cover range is [0 - 100]");
+            }
+            return result;
+        } catch (NumberFormatException e) {
+            errors.rejectValue(CLOUD_COVER, "pl.cyfronet.s4e.data.repository.query.QueryNumber.message", e.getMessage());
+            return 0;
         }
     }
 }
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
@@ -32,24 +32,11 @@ public class ErrorHandlerHelper {
     }
 
     public Map<String, Object> toResponseMap(MethodArgumentNotValidException e) {
-        BindingResult bindingResult = e.getBindingResult();
-        val map = new LinkedHashMap<String, Object>();
+        return getStringObjectMap(e, e.getBindingResult());
+    }
 
-        // populate general errors
-        if (bindingResult.hasGlobalErrors()) {
-            map.put("__general__", bindingResult.getGlobalErrors().stream()
-                    .map(objectError -> getMessage(objectError))
-                    .collect(Collectors.toUnmodifiableList()));
-        }
-
-        // populate field errors
-        for (FieldError fieldError: bindingResult.getFieldErrors()) {
-            map.putIfAbsent(fieldError.getField(), new ArrayList<>());
-            ((List) map.get(fieldError.getField())).add(getMessage(fieldError));
-        }
-
-        optionallyAddDevelopmentInformation(map, e);
-        return map;
+    public Map<String, Object> toResponseMap(QueryException e) {
+        return getStringObjectMap(e, e.getBindingResult());
     }
 
     public Map<String, Object> toResponseMap(RecaptchaException e) {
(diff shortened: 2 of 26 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/data/repository/query/QueryNumber.java
```diff
@@ -13,16 +15,32 @@ public QueryNumber(QueryBuilder queryBuilder) {
...
-            parameters.add(Float.parseFloat(params.get(CLOUD_COVER).toString()));
+            parameters.add(getCloudCover(params, errors));
+        }
+    }
+
+    private float getCloudCover(Map<String, Object> params,
+                                Errors errors) {
+        try {
+            float result = Float.parseFloat(params.get(CLOUD_COVER).toString());
+            if (result < 0 || result > 100) {
```
danielkryska, line 37: What `100` value means?

s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
```diff
@@ -52,6 +52,30 @@
...
+    public Map<String, Object> toResponseMap(QueryException e) {
+        BindingResult bindingResult = e.getBindingResult();
+        val map = new LinkedHashMap<String, Object>();
+
+        // populate general errors
+        if (bindingResult.hasGlobalErrors()) {
+            map.put("__general__", bindingResult.getGlobalErrors().stream()
+                    .map(objectError -> getMessage(objectError))
+                    .collect(Collectors.toUnmodifiableList()));
+        }
```
danielkryska, line 64: Could you move this if to function `populateGeneralErrors`?

s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
```diff
@@ -52,6 +52,30 @@
...
+
+        // populate field errors
+        for (FieldError fieldError : bindingResult.getFieldErrors()) {
+            map.putIfAbsent(fieldError.getField(), new ArrayList<>());
+            ((List) map.get(fieldError.getField())).add(
+                    messageSource.getMessage(fieldError.getCode(),
+                    fieldError.getArguments(),
+                    LocaleContextHolder.getLocale()));
+        }
```
danielkryska, line 73: Could you move this if to function `populateFieldErrors`?

s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
jswk: I would keep it there, as it's not so complicated, and there are only two places it's used in, see "rule of three"

s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
jswk: I would keep it there, as it's not so complicated, and there are only two places it's used in, see "rule of three"

s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
```diff
@@ -52,6 +52,30 @@
...
+                    .collect(Collectors.toUnmodifiableList()));
+        }
+
+        // populate field errors
+        for (FieldError fieldError : bindingResult.getFieldErrors()) {
+            map.putIfAbsent(fieldError.getField(), new ArrayList<>());
+            ((List) map.get(fieldError.getField())).add(
+                    messageSource.getMessage(fieldError.getCode(),
+                    fieldError.getArguments(),
+                    LocaleContextHolder.getLocale()));
```
jswk, line 72: What's wrong with the getMessage method? If you need to pass arguments, you can modify the call in getMessage. `MessageSource::getMessage` is overloaded, and one method has a `@Nullable` args argument.

s4e-backend/src/test/java/pl/cyfronet/s4e/controller/SearchControllerTest.java
```diff
@@ -86,8 +88,38 @@ public void shouldGetSceneBySearchEndpoint() throws Exception {
                 .andExpect(jsonPath("$.length()", is(equalTo(limit))))
                 .andExpect(jsonPath("$[0].artifacts", containsInAnyOrder(
                         "RGB_16b", "RGBs_8b", "checksum", "manifest", "metadata", "quicklook", "product_archive"
-                )))
-        ;
+                )));
+    }
+
+    @Test
+    public void shouldntGetSceneByBadByLimit() throws Exception {
```
jswk, line 95: → shouldntGetSceneByTooLargeLimit?

s4e-backend/src/test/java/pl/cyfronet/s4e/controller/SearchControllerTest.java
```diff
@@ -86,8 +88,38 @@ public void shouldGetSceneBySearchEndpoint() throws Exception {
...
+        return Stream.of(
+                Arguments.of(LIMIT, "abc", "Limit musi być liczbą dodatnią"),
+                Arguments.of(LIMIT, "-1", "Limit musi być liczbą dodatnią"),
+                Arguments.of(OFFSET, "abc", "Offset musi być liczbą dodatnią"),
+                Arguments.of(OFFSET, "-1", "Offset musi być liczbą dodatnią"),
+                Arguments.of(CLOUD_COVER, "abc", "Cloud cover musi być liczbą z zakresu [0 - 100]"),
+                Arguments.of(CLOUD_COVER, "-1", "Cloud cover musi być liczbą z zakresu [0 - 100]"),
+                Arguments.of(CLOUD_COVER, "1000", "Cloud cover musi być liczbą z zakresu [0 - 100]"),
+                Arguments.of(INGESTION_FROM, "2019-11-08", "Zły format daty: `2019-11-08`"),
+                Arguments.of(INGESTION_FROM, "Hakuna matata", "Zły format daty: `Hakuna matata`")
```
jswk, line 121: You've missed the correct parameters here. I'd put them in the same test.
Also, I'd test more thoroughly the edges of the abstraction classes, e.g. cloud cover = -1, 0, 100, 101, NaN, +inf, -inf.

s4e-backend/src/test/java/pl/cyfronet/s4e/data/repository/query/QueryBuilderTest.java
```diff
@@ -62,7 +71,57 @@ public void shouldReturnQueryWithoutAnyConditions() {
...
+                Arguments.of(LIMIT,"abc", true),
+                Arguments.of(LIMIT,"-1", true),
+                Arguments.of(LIMIT,"1000", false),
+                Arguments.of(OFFSET,"abc", true),
+                Arguments.of(OFFSET,"-1", true),
+                Arguments.of(CLOUD_COVER,"abc", true),
+                Arguments.of(CLOUD_COVER,"-1", true),
+                Arguments.of(CLOUD_COVER,"1000", true),
+                Arguments.of(INGESTION_FROM,"2019-11-08", true),
+                Arguments.of(INGESTION_FROM,"Hakuna matata", true)
```
jswk, line 124: You've missed the correct parameters here. I'd put them in the same test.
Also, I'd test more thoroughly the edges of the abstraction classes, e.g. cloud cover = -1, 0, 100, 101, NaN, +inf, -inf.

s4e-backend/src/test/java/pl/cyfronet/s4e/data/repository/query/QueryBuilderTest.java
```diff
@@ -62,7 +71,57 @@ public void shouldReturnQueryWithoutAnyConditions() {
...
+        params.put(INGESTION_FROM, "Hakuna matata");
+        queryBuilder.prepareQueryAndParameters(params, parameters, resultQuery, errors);
+        assertThat(errors.hasErrors(), is(true));
+        assertThat(errors.getFieldError(INGESTION_FROM).getDefaultMessage(),
+                is(equalTo("Text 'Hakuna matata' could not be parsed at index 0")));
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    public void shouldReturnErrorsForParam(String param, Object value, boolean correct) {
```
jswk, line 103: value is always a String in this case

s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
danielkryska: The root of the idea is: put blocks of code in more contextual places - methods - (instead of adding comments like `populate general errors` above them). In this case, the comment wasn't caused by duplication. 

In general: `abstraction should go from top to down`, where detailed method name should describe what they're exactly doing inside. Those methods with comments are a perfect example of it.

s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
danielkryska: The root of the idea is: put blocks of code in more contextual places - methods - (instead of adding comments like `populate general errors` above them). In this case, the comment wasn't caused by duplication. 

In general: `abstraction should go from top to down`, where detailed method name should describe what they're exactly doing inside. Those methods with comments are a perfect example of it.

s4e-backend/src/main/java/pl/cyfronet/s4e/ex/ErrorHandlerHelper.java
jswk: I understand your argument, that one could use language constructs to describe a process. However, I see these constructs also as a hindrance, since you have to switch context to see exactly what is done in a separate method which may be in a different region of the class. Also, making the code conform to some specific requirement may be more difficult, as when you start to reuse the extracted method you have to care about its abstraction.

I agree that _abstraction should go from up to down_, however, it doesn't require you to extract all the tiniest possible abstraction. Instead, weigh pros and cons, in this case I think that level was enough.

danielkryska: I know what you mean saying that there will be `context switch`, but as long as a method is well named (in this case it is) we don't need to dive into the function to understand what it is doing. Let's say that this case isn't `cornerstone` of the PR to discuss it any longer, it is fine as it is. 

PS For the future, mostly I give `approval` with comments that are nice to have for me. Most of them are related to style or placement. If you don't agree with them, just skip it and merge the PR, nothing wrong will happen :)

PR #624, opened 2020-07-27
[#622] Trigger view creation on product modification

jswk:
Before, views were created in SeedProducts, however, this could lead to
inconsistencies, were the product.layer_name modified.


jswk:
> Maybe some tests? ...

This would require new testing metodology for db migrations, might be worth doing, but not in the scope of this ticket

PR #642, opened 2020-08-03
[#641] Create endpoints for existing invitations

### LINKED ISSUE #641: Create endpoints for existing invitations
Source: #483

- [ ] Create endpoints
  - [ ] GET `/invitations`
  - [ ] DELETE `/invitations/{token}`
  - [ ] PUT `/invitations/{token}` - remove old and create new for new email

danielkryska:
- GET invitations
- DELETE inviation by token
- PUT invitation = remove old and send new invitation

Closes #641

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/controller/InvitationController.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/controller/InvitationController.java
@@ -9,16 +9,22 @@ import lombok.RequiredArgsConstructor;
 import lombok.val;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.i18n.LocaleContextHolder;
+import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.bind.annotation.*;
-import pl.cyfronet.s4e.ex.NotFoundException;
-import pl.cyfronet.s4e.controller.request.CreateInvitationRequest;
+import pl.cyfronet.s4e.bean.Invitation;
+import pl.cyfronet.s4e.controller.request.InvitationRequest;
 import pl.cyfronet.s4e.controller.response.BasicInstitutionResponse;
+import pl.cyfronet.s4e.controller.response.BasicInvitationResponse;
 import pl.cyfronet.s4e.event.OnConfirmInvitationEvent;
+import pl.cyfronet.s4e.event.OnDeleteInvitationEvent;
 import pl.cyfronet.s4e.event.OnRejectInvitationEvent;
 import pl.cyfronet.s4e.event.OnSendInvitationEvent;
+import pl.cyfronet.s4e.ex.InvitationInvalidInstitutionException;
+import pl.cyfronet.s4e.ex.NotFoundException;
 import pl.cyfronet.s4e.service.InvitationService;
 
 import javax.validation.Valid;
+import java.util.Set;
 
 import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 import static pl.cyfronet.s4e.Constants.API_PREFIX_V1;
@@ -31,6 +37,18 @@ public class InvitationController {
     private final InvitationService invitationService;
     private final ApplicationEventPublisher eventPublisher;
 
+    @Operation(summary = "Fetch all invitations for specific institution")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Successfully retrieved list"),
+            @ApiResponse(responseCode = "400", description = "Incorrect request", content = @Content),
+            @ApiResponse(responseCode = "401", description = "Unauthenticated", content = @Content),
+            @ApiResponse(responseCode = "403", description = "Forbidden", content = @Content)
+    })
+    @GetMapping(value = "/institutions/{institution}/invitations")
+    public Set<BasicInvitationResponse> get(@PathVariable("institution") String institutionSlug) {
+        return invitationService.findAllBy(institutionSlug, BasicInvitationResponse.class);
+    }
+
     @Operation(summary = "Send invitation email to user")
     @ApiResponses({
             @ApiResponse(responseCode = "200", description = "Invitation was send"),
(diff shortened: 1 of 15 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/config/SecurityConfig.java
```diff
@@ -81,6 +81,8 @@ private void authorizeRequests(ExpressionUrlAuthorizationConfigurer<HttpSecurity
 
                 .mvcMatchers(GET, prefix("/jwt/pubkey")).permitAll()
 
+                .mvcMatchers(DELETE, prefix("/invitation/{token}"))
```
jswk, line 84: Shouldn't it be /invitation**s**/{token}?

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/InvitationController.java
```diff
@@ -40,13 +55,36 @@
...
+    @Operation(summary = "Resend invitation and delete last one")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Invitation was resend"),
+            @ApiResponse(responseCode = "400", description = "Incorrect request", content = @Content),
+            @ApiResponse(responseCode = "401", description = "Unauthenticated", content = @Content),
+            @ApiResponse(responseCode = "403", description = "Forbidden", content = @Content),
+            @ApiResponse(responseCode = "404", description = "Not found", content = @Content)
+    })
+    @PutMapping(value = "/institutions/{institution}/invitation/{token}", consumes = APPLICATION_JSON_VALUE)
+    public void resend(
```
jswk, line 78: Why not return BasicInvitationResponse here? It would make sense for me.
Also, as you're creating and deleting in a single method, I'd annotate it with `@Transactional`, or have a single method in a service which does the two with that annotation.

s4e-backend/src/main/java/pl/cyfronet/s4e/config/SecurityConfig.java
jswk: Also, another thing is that you don't have the institution in the path, so you cannot reference it in the access rule. Either modify the endpoint to be prefixed with `/instititutions/{slug}/` (it would then fall under existing rule) or extend ISH with a checking method such as `isInvitationAdmin` or sth like this. Or `canDeleteInvitation(#token)`, this would hint at entity access model - I would create another class for that, which may later be extended to encompass other entities.

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/response/BasicInvitationResponse.java
```diff
@@ -0,0 +1,15 @@
...
+
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+
+public interface BasicInvitationResponse {
+    public String getEmail();
+
+    public String getToken();
+
+    @Enumerated(EnumType.STRING)
```
jswk, line 13: This isn't needed, as it's not an entity class.

s4e-backend/src/main/java/pl/cyfronet/s4e/service/InvitationService.java
```diff
@@ -92,6 +85,13 @@ public void remove(String token) throws NotFoundException {
         return invitationRepository.findByToken(token, projection);
     }
 
+    public void deleteBy(String token) {
+        val invitation = invitationRepository.findByToken(token, Invitation.class);
+        if (invitation.isPresent()) {
+            invitationRepository.delete(invitation.get());
+        }
+    }
```
jswk, line 93: Shouldn't this throw NFE exception in case invitation is not found? Also, the method should be transactional, as you're doing two operations (find and delete) here.

s4e-backend/src/test/java/pl/cyfronet/s4e/controller/InvitationControllerTest.java
```diff
@@ -128,6 +158,44 @@ public void shouldCreate() throws Exception {
...
+        val URL = API_PREFIX_V1 + "/invitation/{token}";
+        mockMvc.perform(put(URL, invitation.getToken())
+                .contentType(MediaType.APPLICATION_JSON)
+                .with(jwtBearerToken(member, objectMapper))
+                .content(objectMapper.writeValueAsBytes(request))
+        ).andExpect(status().isForbidden());
+    }
+
+    @Test
+    public void shouldResend() throws Exception {
```
jswk, line 175: Apart from this, I would test if an attempt to resend an invitation from another institution fails.
Case would look as follows: two institutions A and B, in B there is an invitation b. institutionAdmin is an admin to A but not B, and he tries to do `PUT /institutions/A/invitation/b` → I think it should effect in 400, as access to A is assured, but there is no match between A and b.

s4e-backend/src/test/java/pl/cyfronet/s4e/controller/InvitationControllerTest.java
```diff
@@ -173,6 +241,17 @@ public void shouldReject() throws Exception {
         assertEquals(invitationAfterReject.get().getStatus(), InvitationStatus.REJECTED);
     }
 
+    @Test
+    public void deleteShouldBeSecured() throws Exception {
```
jswk, line 245: Please verify the happy path of delete as well.

danielkryska:
```
    @Test
    public void shouldSendEmailOnDeletionAndRemoveInvitation() throws NotFoundException {
        val invitation = InvitationHelper
                .invitationBuilder(InvitationHelper.institutionBuilder().build())
                .build();
        when(invitationService.findByToken(invitation.getToken(), Invitation.class))
                .thenReturn(Optional.of(invitation));

        listener.handle(new OnDeleteInvitationEvent(invitation.getToken(), null));

        verify(mailService).sendEmail(eq(invitation.getEmail()), any(), any(), any());
        verify(invitationService).deleteBy(eq(invitation.getToken()));
    }

jswk: This doesn't touch DB and verify authorization.

Issue #648, opened 2020-08-05
Fix circular dependencies while dependency injection


danielkryska:
- [ ] Refactor APP_INITIALIZER
- [ ] Remove all workarounds of InjectorModule usages

jswk:
Hasn't that been completed in a92590d?

jswk:
Closing as no input since last interaction.

PR #649, opened 2020-08-05
[#648] Fix circular dependencies while dependency injection

danielkryska:
- [x] Refactor APP_INITIALIZER
- [x] Remove all workarounds of InjectorModule usages
- [x] Replace all static apiPrefixes with environment variables

Closes: 648

### DIFF
--- a/s4e-web/src/app/utils/initializer/config.service.ts
+++ b/s4e-web/src/app/utils/initializer/config.service.ts
@@ -1,48 +1,42 @@
+import { IRemoteConfiguration } from './../../app.configuration';
 import {Injectable} from '@angular/core';
-import {IConfiguration, IRemoteConfiguration} from '../../app.configuration';
-import {map, switchMap, tap} from 'rxjs/operators';
 import {HttpClient} from '@angular/common/http';
-import {SessionService} from '../../state/session/session.service';
+import environment from 'src/environments/environment';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { tap } from 'rxjs/operators';
+
+@Injectable({providedIn: 'root'})
+export class RemoteConfiguration {
+  public readonly isInitialized$: Observable<boolean>;
+  private _isInitialized$ = new BehaviorSubject<boolean>(false);
+
+  private _configuration: IRemoteConfiguration;
+
+  constructor() {
+    this.isInitialized$ = this._isInitialized$.asObservable();
+  }
+
+  get(): IRemoteConfiguration {
+    return this._configuration;
+  }
+
+  set(configuration: IRemoteConfiguration) {
+    this._configuration = configuration;
+    this._isInitialized$.next(true);
+  }
+}
 
 @Injectable()
-export class S4eConfig implements IConfiguration {
-  backendDateFormat: string;
-  geoserverUrl: string;
-  geoserverWorkspace: string;
-  recaptchaSiteKey: string;
-  timezone: string;
+export class ConfigurationLoader {
+  constructor(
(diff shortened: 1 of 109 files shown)


s4e-web/src/app/views/map-view/state/sentinel-search/sentinel-search.service.spec.ts
```diff
@@ -134,24 +130,24 @@ describe('SentinelSearchResultService', () => {
...
+      store.set([createSentinelSearchResult(SentinelSearchFactory.build(), environment.apiPrefixV1)]);
       service.search({});
       expect(query.getAll().length).toBe(0);
     });
   });
 
   it('openModalForResult', () => {
     const spy = spyOn(modalService, 'show');
-    const result = createSentinelSearchResult(SentinelSearchFactory.build(), apiPrefixV1);
+    const result = createSentinelSearchResult(SentinelSearchFactory.build(), environment.apiPrefixV1);
```
michal-szostak, line 141: Now that `environment.apiPrefixV1` is static (known on compile time) it would be better to remove it from `createSentinelSearchResult` arguments list, and import it there directly.

s4e-web/src/app/utils/initializer/config.service.ts
```diff
@@ -1,48 +1,41 @@
...
-@Injectable()
-export class S4eConfig implements IConfiguration {
-  backendDateFormat: string;
-  geoserverUrl: string;
-  geoserverWorkspace: string;
-  recaptchaSiteKey: string;
-  timezone: string;
+@Injectable({providedIn: 'root'})
+export class RemoteConfiguration {
+  isInitialized$: Observable<boolean>;
```
michal-szostak, line 10: I think it would be more consistent to have method eg.
```
isInitialized$() {
    return this._isInitialized$.asObservable();
}
```

Or at least make this public variable `readonly`;

s4e-web/src/environments/environment.hmr.private.ts
```diff
@@ -5,10 +5,21 @@
...
 export const environment = {
+  //// TODO: Propagate those changes on every environment in base.ts file
+  backendDateFormat: 'YYYY-MM-DDTHH:mm:ss[Z]',
+  generalErrorKey: '__general__',
+  apiPrefixV1: 'api/v1',
+  projection: {toProjection: 'EPSG:3857', coordinates: [19, 52] as [number, number]},
+  maxZoom: 12,
+  timezone: Intl.DateTimeFormat().resolvedOptions().timeZone,
+  ////
+
```
michal-szostak, line 16: I think it's a good moment to move this common environment variables to separate file eg. `environment.common.ts`:

```
# environment.common.ts

export const commonEnvironment = {
    backendDateFormat: 'YYYY-MM-DDTHH:mm:ss[Z]',
    generalErrorKey: '__general__',
    apiPrefixV1: 'api/v1',
    projection: {toProjection: 'EPSG:3857', coordinates: [19, 52] as [number, number]},
    maxZoom: 12,
    timezone: Intl.DateTimeFormat().resolvedOptions().timeZone
}

# every other environment file eg. environment.hmr.private.ts

import sharedEnvironment from './base';
import commonEnvironment from './environment.common';

export const environment = {
version: 'undefined-dev-version',
   ...commonEnvironment
   ...sharedEnvironment,
   production: false,
   hmr: true,
   inviteOnly: true,
   uiDevelopment: true
}

```

This way we can change these parameters in one place, and not break other builds by accident (from what I understand those constants should be the same between builds)

PR #659, opened 2020-08-13
[#621] Add admin Scenes synchronization endpoints

jswk:
The synchronization system is divided into three main parts:
- SyncJobManager: manages jobs state, running operation and details,
- ChunkedRunner: processes a stream of Runnables in chunks,
- SyncJobStore: maintains a list of jobs.

The idea behind using ChunkedRunner is not to load all the scenes keys
into memory, as they can be loaded in chunks while maintaining
parallelism.

During the operation of a job it maintains its state, which is exposed
under GET endpoint - in this way it's possible to monitor the operation
of the job.
For example, it could fail for all keys - in that case it can be
cancelled, and the problem investigated.
All the acceptor Errors are stored in the job for inspection.

The jobs are not persisted to DB, and are the state of a specific
instance of backend.
Memory can be reclaimed by deleting jobs.

Closes: #621.

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/admin/sync/api/AdminSyncJobController.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/admin/sync/api/AdminSyncJobController.java
@@ -1,0 +1,163 @@
+package pl.cyfronet.s4e.admin.sync.api;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.web.bind.annotation.*;
+import pl.cyfronet.s4e.admin.sync.task.SyncJob;
+import pl.cyfronet.s4e.admin.sync.task.SyncJobManager;
+import pl.cyfronet.s4e.admin.sync.task.SyncJobStore;
+import pl.cyfronet.s4e.ex.NotFoundException;
+
+import javax.validation.Valid;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+import static pl.cyfronet.s4e.Constants.ADMIN_PREFIX;
+
+@RestController
+@RequestMapping(path = ADMIN_PREFIX + "/sync-jobs", produces = APPLICATION_JSON_VALUE)
+@Tag(name = "admin-sync-jobs", description = "The Admin Scenes Sync Job API")
+@RequiredArgsConstructor
+@Slf4j
+public class AdminSyncJobController {
+    private final SyncJobStore syncJobStore;
+    private final SyncJobManager syncJobManager;
+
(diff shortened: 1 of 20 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/admin/sync/api/AdminSyncJobController.java
```diff
@@ -0,0 +1,163 @@
...
+
+import javax.validation.Valid;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+import static pl.cyfronet.s4e.Constants.ADMIN_PREFIX;
+
+@RestController
+@RequestMapping(path = ADMIN_PREFIX + "/syncJobs", produces = APPLICATION_JSON_VALUE)
```
danielkryska, line 24: Could you use `css-case` in URL?

s4e-backend/src/main/java/pl/cyfronet/s4e/admin/sync/api/SyncJobMapper.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.stream.Collectors;
+
+@Mapper(config = MapStructCentralConfig.class)
+public abstract class SyncJobMapper {
+    public abstract SyncJobResponse toResponse(SyncJob syncJob);
+
+    public abstract SyncJobResponse.Extended toExtendedResponse(SyncJob syncJob);
+
+    private SyncJobResponse.StateDetailsResponse toResponse(SyncJob.StateDetails stateDetails) {
+        if (stateDetails instanceof SyncJob.PendingDetails) {
```
danielkryska, line 18: Won't be more readable to split this method like this?

```
toResponse() {
  val stateDetails = getDetails(...);
  return stateDetails != null ? toResponse(stateDetail) : null;
}

getDetails() {
switch (typeof stateDetails) {
  case SyncJob.PendingDetails:
    return SyncJob.PendingDetails;
 ... 
  default:
    return null;
}
}

File: s4e-backend/src/main/java/pl/cyfronet/s4e/config/SyncConfig.java [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+import pl.cyfronet.s4e.admin.sync.task.SyncJobStore;
+import pl.cyfronet.s4e.properties.S3Properties;
+import pl.cyfronet.s4e.sync.SceneAcceptor;
+import software.amazon.awssdk.services.s3.S3Client;
+
+import java.time.Clock;
+
+@Configuration
+public class SyncConfig {
+    /// AWS SDK has a default connection pool size of 50, so 40 threads should be fine.
```
danielkryska, line 21: Is this opinion `, so 40 threads should be fine.` necessary?

jswk: You're right, it shouldn't be placed here. I'll externalize this param to a `@ConfigurationProperties` class.

s4e-backend/src/main/java/pl/cyfronet/s4e/admin/sync/api/SyncJobMapper.java
jswk: Unfortunately, it's not possible in Java due to static linking of overloaded methods, see for example https://stackoverflow.com/questions/4523874/is-dynamic-overload-resolution-possible-in-java.
The idea is considered by the MapStruct community though, see issue 131 in their repo (I'm not linking so it doesn't show up in their issue)

PR #664, opened 2020-08-19
[#623] Improve DB normalization

jswk:
Remove s3path and granule_path from table scene, moving their
calculation to a dedicated view: scene_extended.
This view, on the other hand, is used to create views for GeoServer.

A granule_path consists of a prefix and a suffix.
The prefix is now stored as a row in table property, and is initialized
in SeedProducts.
In case no seed were used, admin can modify it's value through endpoints.

The suffix is on the other hand a product of metadata_content,
scene_content and granule_artifact_rule.
It's calculation is moved to a db function, and results go to the
intermediate view scene_extended.

Moreover, timestamp is now a generated field, which is possible as it
only depends on metadata_content.
It is still independently indexed and kept as a column, to allow easier
repository usage.

There is also a cascade of changes in other places, as now scene_content
and metadata_content must always be set.

Closes: #623.


s4e-backend/src/main/java/pl/cyfronet/s4e/util/SceneArtifactsHelper.java
```diff
@@ -0,0 +1,49 @@
...
+            try {
+                return sceneProjection.getSceneContent().get(SCENE_SCHEMA_ARTIFACTS_KEY).get(type).asText().substring(1);
+            } catch (NullPointerException e) {
+                return null;
+            }
+        } else {
+            SceneExtendedProjection sceneExtendedProjection = sceneExtendedRepository.findById(id, SceneExtendedProjection.class)
+                    .orElseThrow(constructNFE(id));
+            return sceneExtendedProjection.getS3Path();
+        }
```
danielkryska, line 43: Should bee better to write it like this

```
if (type == null) {
SceneExtendedProjection sceneExtendedProjection = sceneExtendedRepository.findById(id, SceneExtendedProjection.class)
                    .orElseThrow(constructNFE(id));
            return sceneExtendedProjection.getS3Path();
}

            SceneProjection sceneProjection = sceneRepository.findById(id, SceneProjection.class)
                    .orElseThrow(constructNFE(id));
            try {
                return sceneProjection.getSceneContent().get(SCENE_SCHEMA_ARTIFACTS_KEY).get(type).asText().substring(1);
            } catch (NullPointerException e) {
                return null;
            }

File: s4e-backend/src/main/resources/db/migration/V43__props.sql [hunk 2]
```diff
@@ -0,0 +1,5 @@
+-- Create a table to store properties in, for now it's only the granule_path prefix.
```
danielkryska, line 1: Better name of the `property` table will not be sufficient? What kind of properties will be stored inside this table? AdminProperties? GeneralProperties? GeoSereverProperties? ...

s4e-backend/src/test/java/pl/cyfronet/s4e/admin/property/AdminPropertyControllerTest.java
```diff
@@ -0,0 +1,142 @@
...
+        admin = appUserRepository.save(AppUser.builder()
+                .email("admin@mail.pl")
+                .name("John")
+                .surname("Smith")
+                .password("{noop}password")
+                .enabled(true)
+                .admin(true)
+                .build());
+    }
+
```
danielkryska, line 63: There are missing tests of security checks like
- propertyCreationShouldBeSeecured -> should bee authenticated, should be an admin
...

For each endpoint in controller

danielkryska (changes requested):

s4e-backend/src/main/resources/db/migration/V43__props.sql
jswk: For now it's only a place for the granule_path prefix, which is then incorporated in the scene_extended.granule_path, so for now that's just GeoServerProperties. However, a general abstraction of properties fits here without negative implications, so I wouldn't narrow it's usage in the future down to GeoServer related stuff, as something else may emerge.
Also, I have a feeling, that using a whole table for this is an overkill, but I don't know of an alternative key-value store in postgresql. Maybe you know of something like this (key-value, with RW capabilities from SOK), and we could get rid of the table altogether?

PR #666, opened 2020-08-21
[#541] HTTP code 429 for login and token endpoints

### LINKED ISSUE #541: Add HTTP code 429 (`too many attempts`) and logic for it into API
Steps to reproduce
1. Write email (known)
2. Write password (unknown) 5 times per second (once per 200ms) trying to find it (brute force)

What should happen?
API should return error code 429 `too many attempts`

What happened instead?
Normal requests were returned

kmarszalek:
10 request per 1 minute for /login and /token endpoints

Closes #541


jswk (approved):
The limiting works fine.
Last thing, I'd externalize the dependency versions from `/s4e-backend/pom.xml` to properties in `/pom.xml`.

danielkryska (approved):

Issue #665, opened 2020-08-20
Use admin endpoints instead of SeedProducts


jswk:
Create a script(s), which will work analogously to dataset s4e-sync-1.

The script would have a configurable list of Schemas, Products and prefixes to load.
In this way, it could be used for different datasets, or variants of datasets.

PR #679, opened 2020-09-01
[#665] Add Products seed script

jswk:
**IMPORTANT!!! To review meaningful files in this PR try to run in console:**
```
git checkout 665-admin-endpoints-to-seed;
git diff --no-ext-diff --unified=0 master -- $(git diff --name-only master | grep -vwE "(jpeg|png|tif|scene|zip|metadata|json|md5|xml|txt|gif|BMP|bmp|env|conf|gitignore|list)");
```

The script replaces SeedProducts, integrating in a different lifecycle
moment of the application - it is meant to be run independently of
running specific backend instance.

The seed script operation is configurable from a seed-config: one is
placed in s4e-backend/seed-configs/seed-1.
It is also used in e2e tests, instead of the previous s4e-test-1
dataset, especially when SeedProducts is decommissioned.
Inclusion in e2e tests also ensures it works.

Change in the e2e tests also required few corrections, so that it works
with updated date (now, data in e2e test is from 2020-02-01).

Moreover, configure the default proxy configuration to pass directly to
/api/v1, requiring direct backend access for /api/admin prefix.

Create helper script copy-scenes.sh, which makes it easier to prepare a
bogus data package.
It only copies metadata and scene files, while touching all other
artifacts.

#### Miscellanous

Add local-dataset-1 in minio data directory.

Expose GeoServer seeding endpoints and correct error messages.

Remove SeedProducts and related

Removed properties:
- seed.products.seed-db
- seed.products.sync-geoserver
- seed.products.sync-geoserver.reset-workspace
- seed.products.data-set
- seed.products.s4e-sync-v1.limit
- geoserver.endpoint

Closes: #665.


danielkryska:
I may have a few questions in case of many changes, that can't be fully seen in the Github review.
1. Where are stored Tif's and Images?
2. Why we store metadata in files instead of DB?

jswk:
>     1. Where are stored Tif's and Images?

They are stored in the Ceph buckets, for example in `s4e-sync-1` in `s3.cloud...`.

>     2. Why we store metadata in files instead of DB?

We store it in both, actually. Metadata file is a required artifact of a scene. When the scene file is ingested from the storage, the metadata is copied into the DB and indexed.

danielkryska:
3. On which files should I look at? (There are an enormous stack of them)

jswk:
>     1. On which files should I look at? (There are an enormous stack of them)

It's not that bad really. Filter the changes by commit, so that you don't see the scene and metadata files from the local-dataset-1.

The last commit revolves around `seed-configs/seed-1` and the seed script `seed.sh`. Take a look at them, as they replace preexisting SeedProducts functionality.
Going through the script, you will see new endpoints for seeding GeoServer - these are not explicitly tested, but they use mostly a tested functionality, moreover, gh-actions job `web` tests them indirectly (ensuring `seed.sh` passes and e2e tests pass).
Other than that, there're corrections to nginx config, e2e test, integration test (both depended on specific data from dataset `s4e-test-1`), extension of the `web` gh-actions job and necessary configuration updates.

danielkryska:
> public class AdminGeoserverController
> @PostMapping(path = "/reset-workspace")

Missing responses types handling

danielkryska:
If you could, please rebase your branch against the master, so group removal changes will vanish from git diff... Then I'll continue review

danielkryska (approved):

s4e-backend/src/scripts/seed.sh
```diff
@@ -0,0 +1,215 @@
...
+done
+unset timeout_threshold err
+
+auth_token=$(echo $result | jq -r '.token')
+unset payload result err
+
+temp_file=$(tempfile)
+payload=$(
+  jq -n '$ARGS.named' \
+    --arg value 'mailto://s4e-sync-1/'
```
kmarszalek: maybe a little parametrization :)

kmarszalek (approved):
squash!

PR #688, opened 2020-09-07
[#687] Add Sentry

### LINKED ISSUE #687: Add Sentry

kmarszalek:
Closes #687.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,10 @@
         <unirest.version>3.10.00</unirest.version>
         <bucket4j-starter.version>0.2.0</bucket4j-starter.version>
         <caffeine.version>2.8.5</caffeine.version>
+        <!--
+        update sentry version only when fid instance is updated
+        -->
+        <sentry.version>1.6.3</sentry.version>
     </properties>
 
     <build>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -44,6 +44,7 @@
         <unirest.version>3.10.00</unirest.version>
         <bucket4j-starter.version>0.2.0</bucket4j-starter.version>
         <caffeine.version>2.8.5</caffeine.version>
+        <sentry.version>1.6.3</sentry.version>
```
jswk, line 47: Add a comment here not to update this, unless Sentry instance was updated. I will forget about not upgrading this otherwise

kmarszalek: but i will not! and you need approve :P

jswk: Okay, in such cases I will blame you! :D

s4e-backend/pom.xml
```diff
@@ -42,6 +42,16 @@
...
+                <exclusion>
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-starter-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-log4j2</artifactId>
         </dependency>
```
jswk, line 55: Isn't this a dependency of sentry-log4j2 anyway? You could leave it out as in gs-gateway maybe

jswk (approved):
In general looks good. I'd consider removing this one dependency from s4e-backend.
Moreover, description in README would be nice.

wziajka (approved):

Issue #686, opened 2020-09-04
Prepare RWD version


abacz:
Application doesn't have dedicated RWD version in settings.

PR #691, opened 2020-09-07
[#686] RWD - settings

abacz:


s4e-web/package.json
```diff
@@ -8,7 +8,7 @@
     "start-notifications": "ng build notifications --watch",
     "build-libraries": "ng build notifications",
     "prestart": "npm run make-requirements",
-    "start": "ng serve --proxy-config proxy.conf.json --configuration hmr",
+    "start": "ng serve --host 192.168.1.10 --proxy-config proxy.conf.json --configuration hmr",
```
danielkryska, line 11: Why are you running ng start at this specific host? Are you sure that other developers will correctly run this command?

danielkryska (approved):

s4e-web/package.json
michal-szostak: I'm almost certain that this should not be here...

michal-szostak (approved):

PR #692, opened 2020-09-09
[#671] Migrate seeds of products to specific category and handle them with REST API

### LINKED ISSUE #671: Migrate seeds of products to specific category and handle them with REST API
- [ ] ~~Modify API REST controllers to `/category/{id or label}/product/<...>` type (AdminProductController)~~
  - [x] POST seed product with category
  - [x] PUT change product category by it's name
- [x] Seed [SVG's](https://github.com/cyfronet-fid/sat4envi/files/5181539/icons_categories_svg.zip) into S3 Amazon (example FilesStorage -> upload, delete)
- [x] Generate dynamic icons URLs for seeded categories (See emblems related things. Security -> emblems, mails)
  - [x] From where get prefix to product's categories icons?
  - [x] Icon name should contain extension `.svg`?

```
public String getCategoryUrl(String categoryName) {
    return fileStorageProperties.getKeyPrefixEmblem() + slugify(categoryName);
}
```

- [x] Migrate
  - [x] ~~Category id can be null~~
  - [x] Call createProductCategories() with seed.sh curl
  - [x] Add category migration seeds (defined in [Confluence](https://docs.cyfronet.pl/pages/viewpage.action?pageId=74809551)) and update it's scenes
    - [x] Atmosfera/Meteorologia
      - [x] Zachmurzenie (108m)
      - [x] Chmury konwekcyjne
      - [ ] ~~Wskaźnik stabilności atmosfery~~
    - [x] Morze
      - [ ] ~~Pole wiatru na Morzu Bałtyckim~~
      - [ ] ~~Temperatura powierzchni morza~~
    - [x] Powietrze
      - [ ] ~~Ozon~~
      - [ ] ~~Zawartość dwutlenku azotu~~
      - [x] Pył w atmosferze
    - [x] Deszcze i burze
      - [x] Intensywność opadu
      - [ ] ~~Wilgotność gleby~~ 
      - [x] Obraz satelitarny z nałożonymi wyładowaniami 
      - [x] Suma opadu 
      - [x] Chmury konwekcyjne wysoko wypiętrzone 
    - [x] Powierzchnia
      - [ ] ~~Charakterystyka powierzchni - wysoka rozdzielczość przestrzenna!~~
    - [x] Powierzchnia/Meteorologia
      - [ ] ~~Charakterystyka powierzchni - METEOSAT~~
    - [x] Specjalne
      - [ ] ~~SENTINEL-3, MODIS, METEOSAT - ZAGROŻENIE POŻAROWE~~
      - [ ] ~~SENTINEL-1 - DETEKCJA ZASIĘGU WODY~~
    - [x] ~~Domyślne (null)~~
  - [x] Migrate not classified products to other category
    - Sentinel 1 GRDM
    - Sentinel 1 GRDH
    - Sentinel 1 SLC
    - Airmass
    - Detekcja chmur lodowych i śniegu
    - Mikrofizyka chmur
  - [x] ~~Remove default category~~
  - [x] ~~On null category id add to response default label, name and iconUrl~~
- [x] Tests
  - [x] Create test product should check category
  - [x] Update test product should check category
  - [x] Inititalization of categories should be secured
  - [x] Initialization should add nth categories
  - [x] Should throw not found on non-existing product category when creating or updating product
- [x] Include categories in Products Factory (UI)
- [x] Display icon in this same line as cateogory label

danielkryska:
- Create product with category
- Generate dynamic URLs for category icons
- Add new categories with migration
- Update static products data with categories ids

Closes #671

### DIFF
--- a/s4e-backend/seed-configs/seed-1/products.json
+++ b/s4e-backend/seed-configs/seed-1/products.json
@@ -9,7 +9,8 @@
     "granuleArtifactRule": {
       "default": "quicklook",
       "COG": "RGBs_8b"
-    }
+    },
+    "productCategoryName": "other"
   },
   {
     "name": "Sentinel-1-GRDM",
@@ -20,7 +21,8 @@
     "metadataSchemaName": "Sentinel-1.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "quicklook"
-    }
+    },
+    "productCategoryName": "other"
   },
   {
     "name": "Sentinel-1-SLC_",
@@ -31,7 +33,8 @@
     "metadataSchemaName": "Sentinel-1.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "quicklook"
-    }
+    },
+    "productCategoryName": "other"
   },
   {
     "name": "108m",
@@ -42,7 +45,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "atmosphere-meteorology"
   },
   {
     "name": "Airmass",
@@ -53,7 +57,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "other"
   },
   {
     "name": "Dust",
@@ -64,7 +69,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "air"
   },
   {
     "name": "NatCol",
@@ -75,7 +81,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "other"
   },
   {
     "name": "Polsafi",
@@ -86,7 +93,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "rains-and-storms"
   },
   {
     "name": "RGB24_micro",
@@ -97,7 +105,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "other"
   },
   {
     "name": "Setvak_Eu",
@@ -108,7 +117,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "atmosphere-meteorology"
   },
   {
     "name": "Setvak_PL",
@@ -119,7 +129,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "atmosphere-meteorology"
   },
   {
     "name": "Opad_H03",
@@ -130,7 +141,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "rains-and-storms"
   },
   {
     "name": "Opad_H05_03h",
@@ -141,7 +153,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "rains-and-storms"
   },
   {
     "name": "Opad_H05_06h",
@@ -152,7 +165,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "rains-and-storms"
   },
   {
     "name": "Opad_H05_12h",
@@ -163,7 +177,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "rains-and-storms"
   },
   {
     "name": "Opad_H05_24h",
@@ -174,7 +189,8 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "rains-and-storms"
   },
   {
     "name": "OST",
@@ -185,6 +201,7 @@
     "metadataSchemaName": "MSG.metadata.v1.json",
     "granuleArtifactRule": {
       "default": "product_file"
-    }
+    },
+    "productCategoryName": "rains-and-storms"
   }
 ]
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/admin/product/AdminProductController.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/admin/product/AdminProductController.java
@@ -34,10 +34,10 @@ public class AdminProductController {
             @ApiResponse(responseCode = "403", description = "Forbidden", content = @Content)
     })
     @PostMapping(consumes = APPLICATION_JSON_VALUE)
-    public AdminProductResponse create(@RequestBody @Valid AdminCreateProductRequest request) throws ProductException {
+    public AdminProductResponse create(@RequestBody @Valid AdminCreateProductRequest request) throws ProductException, NotFoundException {
         ProductService.DTO dto = adminProductMapper.toProductServiceDTO(request);
         Long newId = productService.create(dto);
-        return productService.findByIdFetchSchemas(newId, AdminProductResponse.class).get();
+        return productService.findByIdFetchSchemasAndCategory(newId, AdminProductResponse.class).get();
     }
 
     @Operation(summary = "List Products")
@@ -48,7 +48,7 @@ public class AdminProductController {
     })
     @GetMapping
     public List<AdminProductResponse> list() {
-        return productService.findAllFetchSchemas(AdminProductResponse.class);
+        return productService.findAllFetchSchemasAndCategory(AdminProductResponse.class);
     }
 
     @Operation(summary = "Return Product")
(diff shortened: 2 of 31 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/admin/product/AdminProductController.java
```diff
@@ -25,6 +28,70 @@
 public class AdminProductController {
     private final ProductService productService;
     private final AdminProductMapper adminProductMapper;
+    private final ProductCategoryRepository productCategoryRepository;
+
+    @Operation(summary = "Create Product Categories")
```
jswk, line 33: I'd rather move this operation in a separate package `s4e.admin.category` as categories are separate entities.

s4e-backend/src/main/java/pl/cyfronet/s4e/admin/product/AdminProductController.java
```diff
@@ -25,6 +28,70 @@
...
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK"),
+            @ApiResponse(responseCode = "401", description = "Unauthenticated", content = @Content),
+            @ApiResponse(responseCode = "403", description = "Forbidden", content = @Content)
+    })
+    @PostMapping("/categories")
+    public void createCategories() {
+        productCategoryRepository.deleteAll();
```
jswk, line 41: Aren't you deleting the default category by doing this?

s4e-backend/src/main/java/pl/cyfronet/s4e/admin/product/AdminProductController.java
```diff
@@ -40,6 +107,24 @@ public AdminProductResponse create(@RequestBody @Valid AdminCreateProductRequest
...
+    @PatchMapping(path = "/{id}", consumes = APPLICATION_JSON_VALUE)
+    public AdminProductResponse update(
+            @PathVariable Long id,
+            @RequestBody @Valid AdminUpdateProductRequest request
+    ) throws NotFoundException, ProductException {
+        ProductService.DTO dto = adminProductMapper.toProductServiceDTO(request);
+        productService.update(id, dto);
+        return productService.findByIdFetchSchemas(id, AdminProductResponse.class).get();
+    }
+
```
jswk, line 127: What is the rationale for this?

s4e-backend/src/main/java/pl/cyfronet/s4e/bean/Product.java
```diff
@@ -96,7 +96,7 @@
     private Set<AppUser> favourites = new HashSet<>();
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "product_category_id", insertable = false)
+    @JoinColumn(name = "product_category_id", columnDefinition="default 8")
```
jswk, line 99: This will be ignored anyway, as hibernate doesn't create schema in our app.

s4e-backend/src/main/java/pl/cyfronet/s4e/bean/ProductCategory.java
```diff
@@ -24,4 +24,7 @@
 
     @NotEmpty
     private String name;
+
+    @NotEmpty
+    private String iconName;
```
jswk, line 29: Isn't name sufficient in the end?

s4e-backend/src/scripts/seed.sh
```diff
@@ -106,6 +106,17 @@ for schema_path in $(ls $configuration_path/schema); do
...
 done
 
+# Create categories
+echo "Creating categories"
+url=$api_endpoint/api/admin/products/categories
+curl --silent --include --url $url \
+  --request POST \
+  --header "Content-Type: application/json" \
+  --header "Authorization: Bearer $auth_token" \
+  --data "{}"
```
jswk, line 116: I don't think you have to specify this (i.e. `--data "{}"`), other POST requests with an empty request body in this script work fine.

s4e-backend/seed-configs/seed-1/products.json
```diff
@@ -9,7 +9,8 @@
     "granuleArtifactRule": {
       "default": "quicklook",
       "COG": "RGBs_8b"
-    }
+    },
+    "productCategory": 8
```
jswk, line 13: I'd consider using productCategory.name here, you wouldn't need to specify ids in the seeding endpoint

s4e-backend/src/main/java/pl/cyfronet/s4e/admin/product/AdminProductController.java
danielkryska: I was considering seeding the default category inside of REST request due to setting column default with hibernate decorator, but if it doesn't work I will skip default category deletion.

s4e-backend/src/main/java/pl/cyfronet/s4e/bean/ProductCategory.java
danielkryska: No, because of the repetition of used icons

s4e-backend/src/main/java/pl/cyfronet/s4e/service/ProductCategoryService.java
```diff
@@ -0,0 +1,19 @@
...
+public class ProductCategoryService {
+    private final static String ICON_EXTENSION_WITH_DOT = ".svg";
+
+    private final FileStorageProperties fileStorageProperties;
+
+    public String getIconUrlBy(String iconName) {
+        return fileStorageProperties.getKeyPrefixProductsCategoriesIcons()
+                + iconName
+                + ICON_EXTENSION_WITH_DOT;
+    }
```
jswk, line 18: This method should be named `getIconKey`, or something like that, as it's not an url or a path (see SavedViewService for an analogue). Also, `BasicProductCategoryResponse::getUrl` should be rather named `getIconKey` if that's what you want to return. In front-end you then just prepend `/static` and get a valid relative path.

jswk (changes requested):
Thanks for all the corrections. The PR starts looking good.
However, there're a few places with things for correction.

The PR misses the tests for setting the new field in ProductServiceTest. It should be done analogusly to how metadataSchema, sceneSchema or granuleArtifactRule are tested. You should also verify setting the default value (i.e. the default should override current value when updating for example).

I didn't notice this when you introduced the ProductCategory entity, however, there are few compiler warnings you generated:
- `ProductMapper`: you now have `Unmapped target property: "productCategory".` warnings, add the `productCategory` as an ignored property on both mapping methods. The same as other fields are declared to be ignored.
- `Product`, you declared `@EqualsAndHashCode.Exclude` on productCategory, however, it is unnecessary, as on the class you have `@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)`, so you don't have to exclude a field explicitly

Another comiler warnings I noticed by looking at the previous ones are
```
Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
```
which show up at `WMSOverlay.java:12` and `Institution.java:20`. You can safely set the callSuper=false in both cases (these are not related to this PR, but could be improved nevertheless when fixing compiler warnings).

s4e-web/src/app/views/map-view/state/product/product.factory.spec.ts
```diff
@@ -1,5 +1,11 @@
 import * as Factory from 'factory.ts';
-import {Product} from './product.model';
+import { Product, ProductCategory } from './product.model';
+
+const ProductCategoryFactory = Factory.makeFactory<ProductCategory>({
+  id: Factory.each(i => i),
+  label: Factory.each(i => `Product category #${i}`),
+  url: Factory.each(i => `http://test.com/#${i}`),
```
jswk, line 7: Wouldn't the actual values returned by the API be like: `categories/icon_${i}.svg`?

danielkryska: That doesn't matter

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/response/BasicProductCategoryResponse.java
```diff
@@ -1,6 +1,11 @@
 package pl.cyfronet.s4e.controller.response;
 
+import org.springframework.beans.factory.annotation.Value;
+
 public interface BasicProductCategoryResponse {
     Long getId();
     String getLabel();
+
+    @Value("#{@productCategoryService.getIconUrlBy(target.iconName)}")
+    String getUrl();
```
jswk, line 10: I argue, that this field should be named `iconPath` and not `url`.
The value returned by `productCategoryService.getIconUrlBy` is actually something like `categories/icon_sth.svg`, which definitely isn't a [URL](https://en.wikipedia.org/wiki/URL). Even when predending the value with bucket name and getting `static/categories/icon_sth.svg` doesn't fulfill the requirements for a URL, even though clicking such link in context of a browser page would lead you to the resource.
IconPath, on the other hand serves two ends: describes clearly that it is only a path, not a full URL, and describes the content stored at this path, an icon.

@michal-szostak, @kmarszalek? What are your opinions on this, as we cannot get with @danielkryska to understanding.

s4e-backend/src/main/java/pl/cyfronet/s4e/service/ProductService.java
```diff
@@ -98,6 +103,10 @@ public Long create(DTO dto) throws ProductException {
             bindingResult.rejectValue("granuleArtifactRule", GRANULE_ARTIFACT_RULE_MISSING_DEFAULT_CODE);
         }
 
+        val productCategory = productCategoryRepository.findByName(dto.getProductCategoryName(), ProductCategory.class)
+                .orElseThrow(() -> new NotFoundException("Product category with name " + dto.getProductCategoryName() + " doesn't exist"));
```
kmarszalek, line 107: why you dont use constructNFE ?

s4e-backend/src/main/java/pl/cyfronet/s4e/service/ProductService.java
```diff
@@ -111,6 +120,12 @@ public void update(Long id, DTO dto) throws ProductException, NotFoundException
         Product product = productRepository.findById(id)
                 .orElseThrow((() -> constructNFE("Product", id.toString())));
 
+        if (dto.getProductCategoryName() != null && !product.getProductCategory().getName().contains(dto.getProductCategoryName())) {
+            val productCategory = productCategoryRepository.findByName(dto.getProductCategoryName(), ProductCategory.class)
+                    .orElseThrow(() -> new NotFoundException("Product category with name " + dto.getProductCategoryName() + " doesn't exist"));
```
kmarszalek, line 125: same as above, constructNFE?

s4e-backend/src/main/java/pl/cyfronet/s4e/service/ProductService.java
```diff
@@ -98,6 +103,10 @@ public Long create(DTO dto) throws ProductException {
             bindingResult.rejectValue("granuleArtifactRule", GRANULE_ARTIFACT_RULE_MISSING_DEFAULT_CODE);
         }
 
+        val productCategory = productCategoryRepository.findByName(dto.getProductCategoryName(), ProductCategory.class)
```
kmarszalek, line 106: conditions and lines are a little long :)

kmarszalek (changes requested):

kmarszalek (approved):

wziajka (approved):

Issue #683, opened 2020-09-02
[Fix] Add institution with form


danielkryska:
**IMPORTANT!!! Uncomment and run successfully e2e tests in `settings-institution-profile.ts and `settings-institutions-list.spec`**

Steps
I. Add a new institution

1. Go to `/settings/institutions-list` as `zkAdmin@mail.pl`
2. Click `Dodaj instytucję`
3. Fill the form with required fields
4. Save

What should happen?
- User should be redirected to institution profile after creating it
- The institution should be available in institutions search in the sidebar
- The institution should be available in the institutions list

What happens?
- After save user stay on the institution form page
- The institution isn't available in the search
- The institution isn't available in institutions list

II. Add institution child

1. Go to `/settings/institution/<institution-slug>` (any institution available profile) as `zkAdmin@mail.pl`
2. Click `Dodaj instytucję potomną`
3. Fill the form with required fields
4. Save

What should happen?
- User should be redirected to institution profile after creating it
- The institution should be available in institutions search in the sidebar
- The institution should be available in the institution's list
- The institution should be available in the institution's children list of its parent

What happens?
- After save user stay on the institution form page
- The institution isn't available in the search
- The institution isn't available in institutions list
- The institution isn't available in the institution's children list of its parent

UI
- [ ] Run e2e
- [ ] [fix] add/update form
- [ ] [fix] set search value (in `/settings`) as URL instititution
- [ ] Hide from institution member actions buttons when go to institution profile
- [ ] Add return to `mój profil` on institution profile page when visiting as institution member

Backend
- [ ] [fix] Return added/updated institution with response
- [ ] [fix] Update should return actualized institution instead of parent

PR #707, opened 2020-09-14
[#683] fix add institution with form

danielkryska:
UI
- [x] Run e2e
- [x] [fix] add/update form
- [x] [fix] set search value (in `/settings`) as it is in URL `instititution` param

Backend
- [x] [fix] Return added/updated institution with response
- [x] [fix] Update should return actualized institution instead of parent

e2e
- [x] Go to an institution should change the value of institution search and results should be closed

### DIFF
--- a/s4e-backend/src/test/java/pl/cyfronet/s4e/controller/InstitutionControllerIntegrationTest.java
+++ b/s4e-backend/src/test/java/pl/cyfronet/s4e/controller/InstitutionControllerIntegrationTest.java
@@ -12,7 +12,10 @@ import org.springframework.test.web.servlet.MockMvc;
 import pl.cyfronet.s4e.IntegrationTest;
 import pl.cyfronet.s4e.TestDbHelper;
 import pl.cyfronet.s4e.TestResourceHelper;
-import pl.cyfronet.s4e.bean.*;
+import pl.cyfronet.s4e.bean.AppRole;
+import pl.cyfronet.s4e.bean.AppUser;
+import pl.cyfronet.s4e.bean.Institution;
+import pl.cyfronet.s4e.bean.UserRole;
 import pl.cyfronet.s4e.controller.request.CreateChildInstitutionRequest;
 import pl.cyfronet.s4e.controller.request.CreateInstitutionRequest;
 import pl.cyfronet.s4e.controller.request.UpdateInstitutionRequest;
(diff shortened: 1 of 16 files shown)


s4e-web/src/app/views/settings/profile/profile.component.html
```diff
@@ -1,33 +1,85 @@
...
+      data-ut="member-institutions-tile"
     >
-      Zmień hasło
-    </button>
-  </footer>
-</section>
+      <img src="./assets/images/ico_institution.svg" alt="Ikona instytucji" width="50" tile-image />
+      <header class="panel__header" tile-title>Instytucje członkowskie</header>
+      <div tile-description>
+        <p *ngFor="let institution of (memberInstitutions$ | async).slice(0, MAX_INSTITUTIONS_TO_DISPLAY_PER_TILE)">
```
michal-szostak, line 61: This is something that should not be done (calling functions on collections which are iterated in template)
this slice function will be called on every refresh cycle and whole list will be rerendered.
I would suggest adding this slice operation to the pipe of `memberInstitutions$` observable.

s4e-web/src/app/views/settings/profile/profile.component.html
```diff
@@ -1,33 +1,85 @@
...
-      </p>
-    </ng-template>
+    <s4e-tile
+      *ngIf="(administrationInstitutions$ | async).length > 0"
+      data-ut="administrative-institutions-tile"
+    >
+      <img src="./assets/images/ico_institution.svg" alt="Ikona instytucji" width="50" tile-image />
+      <header class="panel__header" tile-title>Instytucje administrowane</header>
+      <div tile-description>
+        <p *ngFor="let institution of (administrationInstitutions$ | async).slice(0, MAX_INSTITUTIONS_TO_DISPLAY_PER_TILE)">
```
michal-szostak, line 31: The same issue as with `memberInstitutions$` below

michal-szostak (changes requested):
I have only one code issue which should be fixed before merge, a minor thing

kmarszalek (changes requested):
not sure how you resolved this issue 'Update should return actualized institution instead of parent' .
but i dont see how it was an issue in backend in the first place :)
backend looks good.
please add check in InstitutionControllerIntegrationTest that you got right slug.

s4e-backend/src/test/java/pl/cyfronet/s4e/controller/InstitutionControllerIntegrationTest.java
```diff
@@ -158,7 +169,15 @@ public void shouldUpdateInstitutionWithEmblem() throws Exception {
                 .contentType(MediaType.APPLICATION_JSON)
                 .content(objectMapper.writeValueAsBytes(request))
                 .with(jwtBearerToken(appUser, objectMapper)))
-                .andExpect(status().isOk());
+                .andExpect(status().isOk())
+                .andExpect(
```
kmarszalek, line 173: .andExpect(jsonPath("$.name",is(equalTo(request.getName()))))

kmarszalek (approved):

PR #711, opened 2020-09-15
[Hotfix] Categories url builder for S3 icons

danielkryska:
what happens?
Generated url is incorrect (double backslash and slash before file extension) static/categories//ico_earth/.svg

What will happen?
Generated url will look like: static/categories/ico_earth.svg

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/service/ProductCategoryService.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/service/ProductCategoryService.java
@@ -7,17 +7,16 @@ import pl.cyfronet.s4e.properties.FileStorageProperties;
 @Service
 @RequiredArgsConstructor
 public class ProductCategoryService {
-    private final static String ICON_EXTENSION_WITH_DOT = ".svg";
+    public final static String ICON_EXTENSION_WITH_DOT = ".svg";
 
     private final FileStorageProperties fileStorageProperties;
 
     public String getIconPath(String iconName) {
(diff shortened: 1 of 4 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/service/ProductCategoryService.java
```diff
@@ -12,12 +12,9 @@
...
 
     public String getIconPath(String iconName) {
-        return String.join(
-                "/",
-                fileStorageProperties.getBucket(),
-                fileStorageProperties.getKeyPrefixProductsCategoriesIcons(),
-                iconName,
-                ICON_EXTENSION_WITH_DOT
-        );
+        return fileStorageProperties.getBucket()
```
kmarszalek, line 15: change delimiter and then it should be good. i wrote test and it was

kmarszalek (changes requested):
add changelog
maybe this will do :) 
`return String.join("", fileStorageProperties.getBucket(), "/", fileStorageProperties.getKeyPrefixProductsCategoriesIcons(), iconName, ICON_EXTENSION_WITH_DOT );`
and also some test, like:
ProductCategoryServiceTest
`@Test public void shouldReturnRightString() { String iconName = productCategoryService.getIconPath("iconName"); assertEquals(iconName, "static/categories-test/iconName.svg"); }`

kmarszalek (approved):

wziajka (approved):

PR #730, opened 2020-09-28
[Fix] E2e optimization and longer timeout

danielkryska:
- turn off video recording
- fix inconsistency in the behavior of .goTo function while sending element to it (dispatching of async tasks in cypress could provide inconsistency)
- remove unnecessary .should('be.visible') to achieve less logs in browser
- clear store before login (force logout on application)
- upgrade of cypress to 4.12.1

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -110,12 +110,6 @@ jobs:
         with:
           name: e2e-results
           path: s4e-web/cypress/screenshots
-      - name: E2E upload videos
-        uses: actions/upload-artifact@v2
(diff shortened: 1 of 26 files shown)


jswk (changes requested):
Seems mostly fine.
However, you missed the `E2E upload videos` step in github actions, which tries to upload the videos when they'll never exist. Remove it to avoid the superfluous operation.

jswk (approved):

michal-szostak (approved):
Rebase and add changelog, apart from that it's good to go

Issue #729, opened 2020-09-28
Assigning products to categories


abacz:
Major:
Assigning products to categories

Minor:
Fix various graphic issues

PR #736, opened 2020-09-30
[#729] Assigning categories & visual fixes

abacz:


michal-szostak (approved):

jswk (approved):
The updated descriptions and categories look fine.

One thing that may be easy to fix is scrolling on the registration page that I noticed:
image
The left part should stay fixed when scrolling. If that's somehow out of scope or requires more plumbing, ignore this.

PR #742, opened 2020-10-02
Backend dependencies update - 2020-10

jswk:
Update postgis docker image to 13-3.0 as new postgresql version has been
released.

Update of justify and joy to new major versions required switching to
jakarta.json-api from javax.json, which is carried out throughout the
code.

Keep current GeoTools version until hibernate-spatial is compatible with
JTS 1.17.x (it introduces a breaking API change).

Make git-commit-id-plugin more forgiving: don't fail if .git directory
isn't available and don't fetch repo information.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,22 +26,25 @@
         <java.version>11</java.version>
         <maven.compiler.release>11</maven.compiler.release>
 
-        <spring-boot.version>2.3.3.RELEASE</spring-boot.version>
+        <spring-boot.version>2.3.4.RELEASE</spring-boot.version>
 
-        <awssdk.version>2.14.7</awssdk.version>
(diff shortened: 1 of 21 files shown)


pom.xml
```diff
@@ -26,22 +26,25 @@
         <java.version>11</java.version>
         <maven.compiler.release>11</maven.compiler.release>
 
-        <spring-boot.version>2.3.3.RELEASE</spring-boot.version>
+        <spring-boot.version>2.3.4.RELEASE</spring-boot.version>
 
-        <awssdk.version>2.14.7</awssdk.version>
+        <awssdk.version>2.15.1</awssdk.version>
```
kmarszalek, line 31: 2.15.2

kmarszalek (approved):

wziajka (approved):

PR #744, opened 2020-10-02
[#743] Add scene key to SearchResponse

jswk:
@michal-szostak: You can display the name part of the key in the front-end search results, instead of the mocked value in a separate ticket.

Closes: #743.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The most recent changes are on top, in each type of changes category.
 
 ### Added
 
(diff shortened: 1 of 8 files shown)


kmarszalek (approved):
it works and looks ok.
we dont want to add some tests for mapper or smth? :)
yeah, i know who should add it in the first place :D

kmarszalek (approved):

michal-szostak (approved):

PR #823, opened 2020-10-23
[Fix] Overlay URL validation

danielkryska:
- Valid localhost
- Valid Http and Https domains
- Valid IP
- Valid subpages
- Valid ports
- Valid params

### DIFF
--- a/s4e-web/src/app/views/map-view/view-manager/overlay-list-modal/overlay-list-modal.component.html
+++ b/s4e-web/src/app/views/map-view/view-manager/overlay-list-modal/overlay-list-modal.component.html
@@ -81,7 +81,7 @@
                   <ng-container *ngIf="hasErrors('url')" >
                     <span
                       class="error"
-                      *ngFor="let error of newOverlayForm.controls.url.errors | keyvalue"
+                      *ngFor="let error of filterWmsUrlErrors(newOverlayForm.controls.url.errors) | keyvalue"
                       [ngSwitch]="error.key"
                       data-e2e="invalid-url-error"
                     >
--- a/s4e-web/src/app/views/map-view/view-manager/overlay-list-modal/wms-url.utils.ts
+++ b/s4e-web/src/app/views/map-view/view-manager/overlay-list-modal/wms-url.utils.ts
@@ -2,19 +2,19 @@ import { AbstractControl } from '@angular/forms';
 
 /* BASE URL */
 const HTTP_PREFIX = '(http(s?):\\/\\/)?';
-const LOCALHOST = '(localhost)';
-const DOMAIN_NAME = '((www\\.)?(([a-zA-Z0-9\\-\\_]*\\.)+)([a-zA-Z0-9\\-\\_]{2,3}))';
-const LOCAL_SUB_PAGES = '((\\/[a-zA-Z0-9]*)+)';
-const IP = '(\\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\b)';
+const LOCALHOST = 'localhost';
+const DOMAIN_NAME = '(www\\.)?((\\.?)[a-zA-Z_\\.-]+)+(\\.[a-zA-Z0-9]{1,3})';
+const LOCAL_SUB_PAGES = '(\\/[a-zA-Z_\\.-]+)';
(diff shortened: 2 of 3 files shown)


s4e-web/src/app/views/map-view/view-manager/overlay-list-modal/wms-url.utils.ts
```diff
@@ -2,19 +2,19 @@ import { AbstractControl } from '@angular/forms';
 
 /* BASE URL */
 const HTTP_PREFIX = '(http(s?):\\/\\/)?';
-const LOCALHOST = '(localhost)';
-const DOMAIN_NAME = '((www\\.)?(([a-zA-Z0-9\\-\\_]*\\.)+)([a-zA-Z0-9\\-\\_]{2,3}))';
-const LOCAL_SUB_PAGES = '((\\/[a-zA-Z0-9]*)+)';
-const IP = '(\\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\b)';
+const LOCALHOST = 'localhost';
+const DOMAIN_NAME = '(www\\.)?((\\.?)[a-zA-Z_\\.-]+)+(\\.[a-zA-Z0-9]{1,3})';
```
jswk, line 6: Domain can have more than 3 chars in the last section.

jswk (changes requested):
The validation now allows a wide range of incorrect inputs:

Incorrect IP:
image

Correct domain (last part can have more than 3 chars) but incorrect protocol:
image

This would lead back to our domain:
image

I don't see use for LOCAL_SUB_PAGES, as you cannot have a WMS service at a local file, like this: file:///some/path.

s4e-web/src/app/views/map-view/view-manager/overlay-list-modal/wms-url.utils.ts
jswk, line 6: Don't handle localhost separately, it should fit in the domain validation. A domain doesn't have to have a dot in it. You can have a valid domain defined in `/etc/hosts` such as `my-great-wms-server`.

Marcelinna (commented):
The validation allows enter incorrect inputs, examples:
aahattpuuu://ss
jj:/s
////s
----/sss
https:/aaaa
htnnnnp://0000
htpps://test/test1/test/2
htpps://0000000/test1/test/2
p1=44&p3=dadadaf
:::////aaaa
::::42000
https://localhost:jjjj
https://localhost:899999999999
llllllocalhostsaaaaa
localhost?0000
90909090
+++9
==2
aaa.ssss.5555
0
00000.0000000000000000
000.0000.0000.0000.000.00

jswk:
> https://localhost:899999999999

Actually, the maximum port number is 65535, see e.g. https://stackoverflow.com/questions/113224/what-is-the-largest-tcp-ip-network-port-number-allowable-for-ipv4.

> llllllocalhostsaaaaa

Isn't that actually a valid domain name? Just default to https and it would be fine probably.

> localhost?0000

This is probably also fine, although requiring an http/https specification might be sensible anyway.

Good catches @Marcelinna!

s4e-web/src/app/views/map-view/view-manager/overlay-list-modal/overlay-list-modal.component.html
```diff
@@ -81,7 +81,7 @@
                   <ng-container *ngIf="hasErrors('url')" >
                     <span
                       class="error"
-                      *ngFor="let error of newOverlayForm.controls.url.errors | keyvalue"
+                      *ngFor="let error of filterWmsUrlErrors(newOverlayForm.controls.url.errors) | keyvalue"
```
michal-szostak, line 84: `filterWmsUrlErrors` should be either a pipe or a stream which is `async` ed.
The methrod here will be called on every update cycle which is an antipattern

s4e-web/src/app/views/map-view/view-manager/overlay-list-modal/wms-url.utils.ts
michal-szostak: I agree

michal-szostak (changes requested):

danielkryska:
After call with @Marcelinna we've come to a conclusion that it's almost impossible to achieve required validation of Web Page. As a page URL can contain numbers and some special chars like for example dots, it's contains definition of IP.

PR #824, opened 2020-10-23
[#808] Give administrator privilidge to user in add insitution form and on people page

### LINKED ISSUE #808: Give administrator privilidge to user in add insitution form and on people page
Only the administrator of a specific institution can do listed actions on /people page

- Add institution comma-separated admins emails while adding/editing institutions (/add-insitution)
- Add institution admin to a person in the list (/people) with checkbox in the column is Admin
- Send an invitation to a user with a possibility to add him admin role by checking box in modal

danielkryska:
Only the administrator of a specific institution can do listed actions on /people page

- Add institution comma-separated admins emails while adding/editing institutions (/add-insitution)
- Add institution admin to a person in the list (/people) with checkbox in the column is Admin
- Send an invitation to a user with a possibility to add him admin role by checking box in modal

Closes #808

### DIFF
--- a/s4e-web/src/app/views/settings/email-list-validator.utils.ts
+++ b/s4e-web/src/app/views/settings/email-list-validator.utils.ts
@@ -1,13 +1,14 @@
 import { AbstractControl } from '@angular/forms';
 
-const COMMA_SEPARATED_EMAILS_EXPRESSION = /((,? ?)((\.?)[a-zA-Z0-9_]+)+@((\.?)[a-zA-Z0-9_]+)+)+/i;
+const COMMA_SEPARATED_EMAILS_EXPRESSION = /^ *((,? *)((\.?)[a-zA-Z0-9_]+)+@((\.?)[a-zA-Z0-9_]+)+)+ *$/i;
 
 interface IValidatorOutput {
   [key: string]: boolean;
 }
 
 export function emailListValidator(control: AbstractControl): IValidatorOutput | null {
-  return _hasMatch(control.value, COMMA_SEPARATED_EMAILS_EXPRESSION) ? null : { emails: true };
+  return !control.value ? null :
+    (_hasMatch(control.value, COMMA_SEPARATED_EMAILS_EXPRESSION) ? null : { emails: true });
 }
 
 function _hasMatch(url: string, regex: RegExp) {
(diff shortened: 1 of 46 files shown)


s4e-web/src/app/views/settings/email-list-validator.utils.ts
```diff
@@ -0,0 +1,15 @@
+import { AbstractControl } from '@angular/forms';
+
+const COMMA_SEPARATED_EMAILS_EXPRESSION = /((,? ?)((\.?)[a-zA-Z0-9_]+)+@((\.?)[a-zA-Z0-9_]+)+)+/i;
```
jswk, line 3: Why not use a ready regex, for example from http://emailregex.com/? The HTML5 W3C version seems quite ok.

jswk (dismissed):

s4e-web/src/app/views/settings/email-list-validator.utils.ts
danielkryska: It's almost the same as `[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+` on-site you just send me, but more human readable (, because without chars in ASCII HEX values).

jswk: Okay, just handle the following cases correctly (input -> result):
"@" -> fail
"d@d" -> ok
" d@d    " -> ok (I'd strip all the whitespace)
" d@d, d@d " -> ok
" d@d, d " -> fail
"d@d,d" -> fail
"d,d@d" -> fail
"s.o_m-e+@mail-0.pl" -> ok
"@d@" -> fail
"d@d@" -> fail
"d@d@d" -> fail
"d" -> fail
"" -> ok (I guess, although I'm not sure about that one)
Make a test for that.

A thing you probably plan to handle is also that it outputs now instead of a sensible message:
image

Marcelinna (commented):
1. Field requires only one email properly, the rest may be entered incorrectly and there is no error.
obraz
obraz


2. The administrators' email field when adding and editing an institution should not be a required field. When trying to save data without filling in this field, the information is displayed: "E-mail or e-mails are incorrect"

3. When checking and unchecking the checkbox in /people, the message "Administrator role for ...... has been removed" is displayed. Even when the checkbox is selected.

4. Removing the user administrator role is not saved. When we remove the roles of the user (uncheck the checkbox) and click the Start link (in the side panel) and then the People link again, the same checkbox is still checked.

michal-szostak (approved):

kmarszalek (approved):

Marcelinna (approved):

michal-szostak (approved):

Issue #830, opened 2020-10-23
Update local-dataset-1 in minio data directory


jswk:
Following the products update from #828 add the corresponding scene stubs in minio resources directory.

PR #831, opened 2020-10-23
[#830] Update local-dataset-1 in minio data directory

jswk:
Add soil moisture and Sentinel 2 stubs.

@wziajka: please create a minio-test image with the updated local-dataset-1 bucket contents, so that no volume mounting is introduced again in the master branch.

Closes: #830.


kmarszalek (changes requested):
i would choose data to have more than one day with more than one time.
Zrzut ekranu 2020-10-23 o 16 49 29

jswk:
> i would choose data to have more than one day with more than one time. other than that it works.. :)

Makes sense, I will add another day worth of data then. @wziajka wait for the update with the image update.

jswk:
@wziajka it's updated now.

kmarszalek (approved):

wziajka (approved):

Issue #835, opened 2020-10-26
Add scientific domain options to AppUser


jswk:
Add the fields from #702 to AppUser.

PR #837, opened 2020-10-26
[#835] Add scientific domain to AppUser

jswk:
Add domain, usage and country fields to app_user table, handle them
during registration and return them in user info.

Make the fields optional for now, so that front-end can be developed in
the meantime.

Closes: #835.


kmarszalek (changes requested):
fill seeds with some data for future testing purpose.

kmarszalek (approved):

michal-szostak (approved):

PR #860, opened 2020-11-05
[#786] OL Loading indicator should only be affected by last OL request

### LINKED ISSUE #786: OL Loading indicator should only be affected by last OL request
## How to reproduce

1. Select product and scene
2. Change scenes fast using next / previous button
3. OL loading indicator disappears, but the scene is loading

## Proposed solution

OL handlers for starting / ending request should be context aware, and context should be reset when new request is made.
(loading finish handler should set loading to false only for the last request)

danielkryska:
- Change Image Wms to Tile Wms loading handling

Closes #786

### DIFF
--- a/s4e-web/src/app/views/map-view/state/overlay/overlay.model.ts
+++ b/s4e-web/src/app/views/map-view/state/overlay/overlay.model.ts
@@ -1,6 +1,9 @@
 import {Layer, Tile} from 'ol/layer';
 import {IUILayer} from '../common.model';
 import {EntityState} from '@datorama/akita';
+import { getBaseUrlAndParamsFrom } from '../../view-manager/overlay-list-modal/wms-url.utils';
+import { NgxUiLoaderService } from 'ngx-ui-loader';
+import {TileLoader} from '../utils/layers-loader.util';
 
 export const GLOBAL_OWNER_TYPE = 'GLOBAL';
 export const INSTITUTIONAL_OWNER_TYPE = 'INSTITUTIONAL';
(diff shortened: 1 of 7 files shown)


s4e-web/src/app/views/map-view/state/overlay/overlay.model.ts
```diff
@@ -61,33 +61,14 @@ export function convertToUIOverlay(
...
-  InjectorModule
-    .Injector
-    .get(NotificationService)
-    .addGeneral({
-      type: 'error',
-      content: 'Wczytanie sceny nie powiodło się'
-    });
+  const tilesLoaderLoader = new TilesLoaderLoader(source);
+  tilesLoaderLoader.start$.then(() => loaderService.startBackground());
+  tilesLoaderLoader.end$.then(() => loaderService.stopBackground());
```
jswk, line 69: From my view, once this typo (double loader) is fixed then the PR is good to go

jswk (changes requested):

michal-szostak:
I have a question - does this PR solves the issue which it is connected to?
Because I indeed see change from `Image` to `Tile`, but I do not see anything which would solve the #786.
(The PR instance is down, so I couldn't check there)

danielkryska:
@michal-szostak I belive this was the issue. 

Before we were using `OL` image hooks like `imageloadend` and `imageloadstart` for `ImageWms` usage only.
```
    source.once('imageloadend', () => resolve());
    source.once('imageloadstart', () => resolve());
```

, but then @jswk changed handling to `Tiles`, so finally loading methods names should be changed also. In this case to `tileloadend` and `tileloadstart`. Loading indicator we've seen `sometimes` while choosing `product` was misleading and the bug wasn't connected with unsloved stream/promise/async.
 
```
    this.end$ = new Promise((resolve) => source.once('tileloadend', () => resolve()));
    this.start$ = new Promise((resolve) => source.once('tileloadstart', () => resolve()));

danielkryska:
@michal-szostak I've updated handling of tiles loader to switch map to new incoming promise. It's working much smoother than before.

jswk (approved):
Now it works correctly. Thanks for fixing this :)

michal-szostak (approved):

Issue #862, opened 2020-11-06
Add ZK option to Institution


jswk:
Add a boolean field `zk` to Institution and propagate it in the hierarchy.

The field should be settable by admin. Allow only to unset it on the root of ZK (that is on an institution which has no parent or its parent has `zk==false`).

Use it to populate `ROLE_ZK_MEMBER`. Remove obsolete `AppUser.zkMember`.

kmarszalek:
leave zkMember flag on user/me endpoint

PR #871, opened 2020-11-10
[#862] Add ZK option to Institution

jswk:
Add a boolean field zk to Institution and propagate it in the hierarchy.

The field should be settable by admin in AdminInstitutionController.
Allow only to unset it on the root of ZK (that is on an institution
which has no parent or its parent has zk==false).

Use it to populate ROLE_ZK_MEMBER.
Remove obsolete AppUser.zkMember.

Closes: #862.


s4e-backend/src/main/java/pl/cyfronet/s4e/admin/institution/AdminInstitutionController.java
```diff
@@ -0,0 +1,50 @@
...
+            @ApiResponse(responseCode = "403", description = "Forbidden", content = @Content)
+    })
+    @PatchMapping(path = "/{slug}/zk/set")
+    public void setZk(@PathVariable String slug) throws NotFoundException, InstitutionZkException {
+        institutionService.setZk(slug);
+    }
+
+    @Operation(summary = "Unset ZK flag")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK"),
```
danielkryska, line 41: Wouldn't be better to response with message like `zk flag was unset`?

s4e-backend/src/main/java/pl/cyfronet/s4e/admin/institution/AdminInstitutionController.java
```diff
@@ -0,0 +1,50 @@
...
+@RestController
+@RequestMapping(path = ADMIN_PREFIX + "/institutions", produces = APPLICATION_JSON_VALUE)
+@Tag(name = "admin-institution", description = "The Admin Institution API")
+@RequiredArgsConstructor
+public class AdminInstitutionController {
+    private final InstitutionService institutionService;
+
+    @Operation(summary = "Set ZK flag")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK"),
```
danielkryska, line 29: Wouldn't be better to response with message like `zk flag was set`?

s4e-backend/src/main/java/pl/cyfronet/s4e/service/InstitutionService.java
```diff
@@ -29,65 +25,91 @@
...
 
     @Transactional(rollbackFor = InstitutionCreationException.class)
-    public Institution save(Institution institution) throws InstitutionCreationException {
-        try {
-            return institutionRepository.save(institution);
-        } catch (DataIntegrityViolationException e) {
-            log.info("Cannot create Institution with name '" + institution.getName() + "'", e);
-            throw new InstitutionCreationException("Cannot create Institution", e);
-        }
+    public String create(Institution institution) throws InstitutionCreationException {
```
danielkryska, line 44: why this function have been renamed from `save` to `create`? Haven't used before `save` name of such methods?

s4e-backend/src/test/java/pl/cyfronet/s4e/TestJwtUtil.java
```diff
@@ -68,9 +68,9 @@ private static AppUserDetails createAppUserDetails(AppUser user) {
         if (user.isAdmin()) {
             authorities.add(new SimpleGrantedAuthority("ROLE_ADMIN"));
         }
-        if (user.isMemberZK()) {
-            authorities.add(new SimpleGrantedAuthority("ROLE_MEMBER_ZK"));
-        }
+//        if (user.isMemberZK()) {
```
danielkryska, line 71: Is the commented code necessary?

danielkryska (approved):

kmarszalek (approved):

s4e-backend/src/main/java/pl/cyfronet/s4e/service/InstitutionService.java
jswk: There are some methods with save, although most of the creational methods (among services) stem from create. Also, the child-creating method in InstService already had the create in name, so I homogenized the naming.

PR #876, opened 2020-11-13
[Fix] User details in profile

danielkryska:
- Chaneg static name and surname to dynamic one using session store


kmarszalek (approved):

jswk (approved):
Maybe since you're adding profile fields you could also expose domain, usage and country input during registration? (See AppUserResponse.)

Issue #794, opened 2020-10-19
[Fix] view should load product scene


Marcelinna:
Tested on instance https://sok.grid.cyfronet.pl
 SOK version v13.2.0

**Steps:**
1. Log in ZkAdmin account
2. Choose an image from the product (e.g Airmass)
3. Click sending a view ("Wyślij widok jako wiadomość")
4. Click the link to view
![obraz](https://user-images.githubusercontent.com/59872096/96516790-0bdf1480-1268-11eb-83cf-4e408d5c8b2d.png)

**What happens?**
The chose image is not loading.

**What was expected to happen?**
The chose image is loading right after the open the link.

The same situation for display saved views

**Steps:**
1. Log in ZkAdmin account
2. Choose an image from the product (e.g Airmass)
3. Add layer "gminy"
4. Click link "Zapisz widok"
4. Click link "Wyświetl zapisane widoki"

**What happens?**
The chose image is not loading, the user sees only layer "gminy"

**What was expected to happen?**
The chose image is loading right after the open the link.

Marcelinna:
similar problem Products do not load right away #817

PR #897, opened 2020-11-17
[#794] [Fix] View should load scene

danielkryska:


s4e-web/src/app/views/map-view/state/map/map.query.ts
```diff
@@ -28,7 +29,8 @@ export class MapQuery extends Query<MapState> {
         overlays: overlays.length === 0 ? null : overlays,
         product: product == null ? null : product.active,
         scene: scene == null ? null : scene.id,
-        date: product.ui.selectedDate,
+        date: !!scene && moment(scene.timestamp).format('YYYY-MM-DD[T]HH:mm:ss[Z]')
+          || moment(product.ui.selectedDate).format('YYYY-MM-DD'),
```
jswk, line 33: Shouldn't a date be always in `YYYY-MM-DD` format? After all `YYYY-MM-DD[T]HH:mm:ss[Z]` is not a date anymore, but a timestamp/date-time.
I don't mean that it's erroneous but I'm trying to understand why would that make sense, as the original was of setting it by means of `product.ui.selectedDate` seems more sensible to me at the moment.

PR #912, opened 2020-11-20
[Fix] Scroll page position to top on navigate in settings

danielkryska:


s4e-web/src/app/views/settings/settings.component.ts
```diff
@@ -60,6 +60,19 @@ export class SettingsComponent implements OnInit, OnDestroy {
...
       });
 
+    // Set page scroll at initial place
+    window.scrollTo(0, 0);
+    this._router.events
+      .pipe(
+        filter(event => event instanceof NavigationEnd),
+        map(() => this._router.url),
+        filter(url => url.indexOf('/settings') > -1),
+        tap(url => console.log(url)),
```
jswk, line 70: I'd remove this log.

jswk (approved):
I tested it locally and it works correctly.
Apart from the log removal it is fine for merging.

s4e-web/src/app/views/settings/settings.component.ts
michal-szostak: I concur

s4e-web/src/app/views/settings/settings.component.ts
```diff
@@ -60,6 +60,19 @@ export class SettingsComponent implements OnInit, OnDestroy {
         }
       });
 
+    // Set page scroll at initial place
+    window.scrollTo(0, 0);
+    this._router.events
+      .pipe(
+        filter(event => event instanceof NavigationEnd),
+        map(() => this._router.url),
```
michal-szostak, line 68: I believe it's better to use url provided by `event`

michal-szostak (approved):
I have 2 things to improve, but generally it's good to go

Issue #919, opened 2020-11-23
Add Institution deleter role


jswk:
Add `OP_INST_DELETE` and `OP_INST_DELETE_GRANT` authorities for deleting managed institutions and granting the former.

PR #926, opened 2020-11-24
[#919] Add authority allowing to delete managed Institutions

jswk:
Add endpoints to grant authorities.

Move authorities such as admin or eumetsat_license to
AppUser.authorities, so that they are settable dynamically.

Closes: #919.


danielkryska (approved):

kmarszalek (changes requested):
ids to mails, info via slack.

kmarszalek (approved):

Issue #903, opened 2020-11-18
Logged in user can navigate to /login


michal-szostak:
On instance without persisting store user can load `/login` page (which he shouldn't as loading `profile` is done on `APP_INITIALIZERS`)

PR #930, opened 2020-11-24
[#903] Prevent logged in user to access /login page

michal-szostak:
Closes: #903

### DIFF
--- a/s4e-web/cypress/integration/auth/auth-login.spec.ts
+++ b/s4e-web/cypress/integration/auth/auth-login.spec.ts
@@ -4,7 +4,7 @@ import { Login } from '../../page-objects/auth/login.po';
 
 
 describe('Auth', () => {
-  
+
   before(() => {
     cy.fixture('users/zkMember.json').as('zkMember');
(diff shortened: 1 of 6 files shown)


s4e-web/cypress/integration/auth/auth-login.spec.ts
```diff
@@ -55,6 +55,11 @@ describe('Auth', () => {
       Login
         .fillForm({ ...this.userToRegister, email: 'wrongFormat' })
         .errorsCountShouldBe(1);
-    })
+    });
+
+    it('should not allow login page for user with session', function () {
+      Login.loginAs(this.superAdmin)
+      cy.visit('/login')
```
danielkryska, line 62: Shouldn't be checked if user is redirected to `/map/products`?

danielkryska (approved):

Marcelinna (approved):

s4e-web/cypress/integration/auth/auth-login.spec.ts
michal-szostak: Right, I'll fix it

PR #933, opened 2020-11-25
Fix overlaping overlays

danielkryska:
- Products on overlays div resize should be always visible

image


kmarszalek (approved):
it works.

michal-szostak (approved):
It would be nice to have some kind of test for that, but I understand that these kind of features are hard to test.

Unit test of `resize` / `mimicElement` itself would not assure that the feature as a whole is working as intended.

Issue #924, opened 2020-11-24
Implement WCAG High Contrast


abacz:
Prevs:
https://invis.io/EBZISQQYK98
https://invis.io/N6ZISR1RYQV
https://invis.io/S9RJNE8WXUB#/434415722_WCAG_-_Modal
https://invis.io/S9RJNE8WXUB#/434415723_WCAG_-_Sentinel_Search
https://invis.io/S9RJNE8WXUB#/434415724_WCAG_-_Ustawienia

abacz:
I didn't use this prevs. After discussion with BW we admited that using filter: invert() will be quicker + some coloristic improvements.

PR #935, opened 2020-11-25
[#924] WCAG High contrast

abacz:


danielkryska (approved):

michal-szostak (approved):

Marcelinna:
it is not possible to turn off high contrast

PR #938, opened 2020-11-25
[#702] Add domain, usage, country and policy to registration form

danielkryska:


s4e-web/src/app/views/register/state/register.model.ts
```diff
@@ -4,8 +4,62 @@ export interface RegisterFormState {
...
 }
 
 // tslint:disable-next-line:no-empty-interface
 export interface RegisterState {}
+
+export const scientificDomainsTypes = [
+  {
+    label: 'Atmosfera',
+    value: 'ATMOSPHERE'
+  },
```
michal-szostak, line 21: These constants should be of some type - probably something like 
```
interface Option<Val> {
label: string;
value: Val
}

DomainType = 'ATMOSPHERE' | 'MARINE' ...
```

s4e-web/src/app/views/register/state/register.model.ts
```diff
@@ -4,8 +4,62 @@ export interface RegisterFormState {
...
+    label: 'Inne',
+    value: 'OTHER'
+  }
+];
+
+export const appUsageTypes = [
+  {
+    label: 'Przeprowadzam badania',
+    value: 'RESEARCH'
+  },
```
michal-szostak, line 52: same as with `scientificDomainsTypes`

michal-szostak (approved):
Some minor code wise things could be fixed, but generally it's ok

jswk (approved):

PR #946, opened 2020-11-26
Change user profile view

danielkryska:
- Add section `remove user`

### DIFF
--- a/s4e-web/src/app/state/form/form.model.ts
+++ b/s4e-web/src/app/state/form/form.model.ts
@@ -20,4 +20,5 @@ export interface FormState {
   invitation: InvitationForm;
   jwtToken: JwtTokenForm;
   expertHelp: ExpertHelpForm;
+  removeUser: {password: string};
 }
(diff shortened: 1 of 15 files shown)


kmarszalek:
in section 'Usuwanie konta' there is a button 'zmień hasło'? 

Zrzut ekranu 2020-11-26 o 09 12 58

jswk (changes requested):
by @kmarszalek:
> in section 'Usuwanie konta' there is a button 'zmień hasło'?
> 
> Zrzut ekranu 2020-11-26 o 09 12 58

@danielkryska re-request review once the PR is ready

jswk (changes requested):
1. The password input in removeUser should only be validated to be not empty, but not require minimum length.
2. When input password is invalid modal shouldn't show.
3. Currently when password is invalid, then I'm logged out, instead of just getting info that password is invalid.

michal-szostak (changes requested):

jswk (approved):
All the points (1,2,3) from my previous review are solved, thanks!

I'd expect also to see a notification that removal was complete on being redirected to /login after successful operation, but it's non-blocking for me.

jswk:
Since you're also modifying register page, I'd add a back-link to the main view there. Analogously as there is one in login view.

michal-szostak (approved):

PR #952, opened 2020-11-27
Add institutional overlays management

danielkryska:
- Add manage overlays button into `/settings/institution?institution=...` (institution profile)


s4e-web/src/app/components/overlay-list/overlay-list.component.html
```diff
@@ -0,0 +1,241 @@
+<ng-container *ngIf="!(isLoading$ | async); else loadingRef">
+  <div *ngIf="!(overlays$ | async) || (overlays$ | async).length === 0" class="infobox infobox--emptylayers">
+    <p>
+      Nie masz jeszcze żadnej swojej nakładki. <a href="#">Sprawdź czym są nadkładki.</a>
```
michal-szostak, line 4: Where does this link lead?

michal-szostak (approved):

s4e-web/src/app/components/overlay-list/overlay-list.component.html
danielkryska: I don't know... I belive Andrzej would know

Marcelinna:
It is not possible to add two layers with the same name because I get error 404

danielkryska:
> It is not possible to add two layers with the same name because I get error 404

It's backend stuff

Marcelinna:
Although admin@mail.pl in Global WMS will turn off the visibility of layers, they are visible in all accounts

Marcelinna:
After adding layer, it isnt visible 
image

Marcelinna:
> Although [admin@mail.pl](mailto:admin@mail.pl) in Global WMS will turn off the visibility of layers, they are visible in all accounts

This same situation for user with administrator privileges when they turn off layer for layer in /settings/institution-wms-overlays?

jswk (approved):
Global WMS shouldn't be exposed in the UI. They are to be API managed, as you need to reconfigure several components when modifying them.

kmarszalek (approved):

Issue #968, opened 2020-11-30
[Fix] The occurrence of the "Osoby w instytucji", "Warstwy WMS dla instytucji" element for a regular member without administrator privileges


Marcelinna:
Tested on instance SOK version master-f9c7b4dc272d0d8e1bdcf028a8ebdec08682bf036

Steps:
1. Log as zkMember
2. Go to /settings/institution?institution=zarzadzenie-kryzysowe-pl

What happens?
![image](https://user-images.githubusercontent.com/59872096/100616180-04a42f80-3319-11eb-8a4a-a9f506200471.png)
What was expected to happen?
User without administrator privileges should not see "Osoby w instytucji"

Marcelinna:
works on SOK v 16.1.0

Issue #787, opened 2020-10-19
During registration, the information that the password should contain at least 8 characters is not displayed


Marcelinna:
Tested on instance https://sok.grid.cyfronet.pl
SOK version v13.2.0

Steps:

1. Go to /register
2. Enter less than 8 characters in the password field

What happends?
The information isn't displayed "Hasło powinno mieć co najmniej 8 znaków"

What was expected to happend?
The information is displayed "Hasło powinno mieć co najmniej 8 znaków"

Marcelinna:
work on SOK version master-d04728dd58130718cce8f68d9a8944a69d5d4835

PR #972, opened 2020-12-01
[Fix] registartion fields, remove breadcrumb in institution profile, hide add global wms

danielkryska:
- [x] During registration, the information that the password should contain at least 8 characters is not displayed #787
- [x] [Fix] Registration valid policy confirmation
- [x] [Fix] The occurrence of the "Osoby w instytucji" element for a regular member without administrator privileges #968
- [x] Hide add global wms from admin@mail.pl account


kmarszalek (approved):
it works as intended

Marcelinna:
"Osoby w instytucji" element is hide also for member with administrator privileges

jswk (approved):

PR #973, opened 2020-12-01
Fixes for: drawing polygon, institution form, admin dashboard

danielkryska:
- Skip s3 image loading error (empty response or 404 response) in institution form
- [Fix] Disable parent selection in institution edit mode and add institution child mode
- Remove `add_child` query param from URL on add child institution
- [Fix] Disable polygon drawing when products tab is on (should work only in sentinel search mode)
- [Fix] Multiple institutions admin is redirected to institutions list when go to /settings URL
- Remove breadcrumbs functionality from code
- [Fix] Inconsistent navigation in settings (landing on different pages)


kmarszalek:
if you load product and then go to search you can draw polygon on loaded product. but if you are on product list you cant.so i would leave it as a bug?
Zrzut ekranu 2020-12-1 o 15 15 51

kmarszalek:
why dont you unable the field with parent in while editing institution, now i can choose diff parent and save but my parent wont change..

kmarszalek (changes requested):
as described in previous comments

kmarszalek (approved):
should work 👍

Marcelinna (approved):

kmarszalek (changes requested):
found one, you can still edit parent institution. even if it doesnt change anything on save.

michal-szostak (approved):

danielkryska:
> found one, you can still edit parent institution. even if it doesn't change anything on save.

I don't know which browser do you use or what state of store do you have, but the input is disabled on edition and adding child. Unfortunately there is an other bug that can cause opening modal. Label click call function that allow change of the institution parent.

This issue will be fixed in other PR.

kmarszalek (approved):

PR #985, opened 2020-12-03
Resize institution emblem on upload

danielkryska:
image

- Allow user to upload files with larger images then 500 x 500 and lesser then 130 x 130 by scaling then up and down


michal-szostak (approved):

jswk (approved):

kmarszalek (approved):

Marcelinna:
image
Logo goes beyond the area of ​​the institution

Marcelinna:
if the width and height are not the same size, error 500 is dispalyed

Marcelinna:
Lack of validation for other extensions, when user put svg file, error 500 is displayed

Issue #986, opened 2020-12-03
Handle PAK role


jswk:
After adding a separate role for PAK users (see #984) handle them accordingly in the menu.

They should have the same visibility of functions in the menu as ZK, save for the expert-mail.

PAK members have an extra authority in the /users/me output: `ROLE_MEMBER_PAK`.

PR #997, opened 2020-12-04
[#986] Show zk memebr functionalities to pak user

danielkryska:
- show funcitonalitites in map dropdown to pak user

### DIFF
--- a/s4e-web/src/app/state/session/session.query.ts
+++ b/s4e-web/src/app/state/session/session.query.ts
@@ -35,6 +35,11 @@ export class SessionQuery extends Query<Session> {
     return this.getValue().admin;
   }
 
+  selectPakMember() {
+    return this.select('authorities')
+      .pipe(map(authorities => authorities.some(authority => authority === 'ROLE_MEMBER_PAK')));
+  }
+
   canDeleteInstitution() {
     return this.getValue().authorities.includes('OP_INSTITUTION_DELETE') || this.getValue().admin;
   }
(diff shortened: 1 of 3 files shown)


s4e-web/src/app/state/session/session.query.ts
```diff
@@ -35,6 +35,11 @@ export class SessionQuery extends Query<Session> {
     return this.getValue().admin;
   }
 
+  selectPakMember() {
+    return this.select('authorities')
+      .pipe(map(authorities => authorities.some(authority => authority === 'ROLE_MEMBER_PAK')));
```
michal-szostak, line 40: use `authorities.includes('ROLE_MEMBER_PAK')`

s4e-web/src/app/views/map-view/map-view.component.ts
```diff
@@ -83,7 +83,10 @@ export class MapViewComponent implements OnInit, OnDestroy {
   public showProductDescription$: Observable<boolean> = this.mapQuery.selectShowProductDescription();
   public selectedLocation$: Observable<LocationSearchResult | null>;
   public overlays$: Observable<UIOverlay[]> = this.overlayQuery.selectVisibleAsUIOverlays();
-  public userIsZK$: Observable<boolean> = this.sessionQuery.selectMemberZK();
+  public userIsAuthorizedForAdditionalFunctionalities$: Observable<boolean> = forkJoin([
+    this.sessionQuery.selectPakMember(),
+    this.sessionQuery.selectMemberZK()
+  ]).pipe(mapAnyTrue());
```
michal-szostak: `forkJoin` is not appropriate here - use something like `combineLatest`.
`forkJoin` emits value when all it's observable finishes, which will not occur in this case.

(`forkJoin` is good for combining http requests which completes)

michal-szostak (approved):

jswk (approved):

PR #1008, opened 2020-12-09
[#1006] Set domain for authorization cookie

jswk:
Depending on the property jwt.cookie.domain the cookie's domain is
either fetched from header Host (in case the property is null or blank)
and set to the value from the property otherwise.

Closes: #1006.


kmarszalek (approved):

danielkryska (approved):

kmarszalek (changes requested):
dot

kmarszalek (approved):

Issue #1017, opened 2020-12-11
Set From mail header


jswk:
Set From mail header, without which some SMTP servers refuse to send emails.

PR #1018, opened 2020-12-11
[#1017] Set From outgoing mail header

jswk:
Populate the header's value based on Spring MailProperties, so it is
consistent with auto-configuration of the JavaMailSender.

Leaving the field empty causes some SMTP servers to reject such requests.

Fixes: #1017.


kmarszalek (approved):
there are more instances of sending

wziajka (approved):

PR #1033, opened 2020-12-17
Fix e2e tests

Marcelinna:
- fixed e2e map test
- added user-options tests
- added map-data-search tests


s4e-web/cypress/page-objects/map/map-locations-search.po.ts
```diff
@@ -60,17 +59,17 @@ export class LocationsSearch extends Core {
...
     LocationsSearch
       .pageObject
       .getSearchResults()
       .eq(nth)
-      .click({ force: true });
+      .click();
 
     return LocationsSearch;
-  }
+  };
```
danielkryska, line 72: Semicolon after methods isn't required. Please remove it due to consistency in code (not only cypress files).

s4e-web/cypress/page-objects/map/map-locations-search.po.ts
```diff
@@ -89,31 +88,31 @@ export class LocationsSearch extends Core {
...
       .getSelectActiveResultBtn()
-      .click({ force: true });
+      .click();
 
     return LocationsSearch;
-  }
+  };
 
-  static waitForCitiesList() {
+  static waitForCitiesList(name:string) {
```
danielkryska, line 111: You could merge methods `type` and `waitorCitiesList` into one `searchCitiesBy(searchValue: string)`

danielkryska (changes requested):

s4e-web/cypress/page-objects/map/map-locations-search.po.ts
Marcelinna: I use .type() for type {enter} key too so I can't merge it

Issue #1034, opened 2020-12-17
Homogenize spatial axis order


jswk:
Currently, in different places there is varying handling of coordinate order.
In search endpoints lat/long is expected as input, but long/lat is returned.
In SearchResponse on the other hand the returned order is lat/long.

WKT uses long/lat order and it should be more straightforward to switch to, it is also how DHuS handles input parameters.

PR #1035, opened 2020-12-17
[#1034] Homogenize lat/long order

jswk:
In endpoints always return and require coordinates in EPSG:4326 lat/long
order.
Update apihowto to include this.

Internally, leave long/lat in some cases (metadata for example).

Always return coordinates with 5 significant digits: it corresponds to
around 1m of precision, which is more than necessary in our case.

Improve parsing of geometry params in OSearch API.

Fixes: #1034.

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/data/repository/query/QueryBuilderImpl.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/data/repository/query/QueryBuilderImpl.java
@@ -15,7 +15,7 @@ public class QueryBuilderImpl implements QueryBuilder {
         columns.add(Scene.COLUMN_ID);
         columns.add(Scene.COLUMN_PRODUCT_ID);
         columns.add(Scene.COLUMN_SCENE_KEY);
-        columns.add("ST_AsText(ST_Transform(" + Scene.COLUMN_FOOTPRINT + ",4326)) AS footprint");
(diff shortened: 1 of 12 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/data/repository/query/QueryBuilderImpl.java
```diff
@@ -15,7 +15,7 @@ public QueryBuilderImpl() {
         columns.add(Scene.COLUMN_ID);
         columns.add(Scene.COLUMN_PRODUCT_ID);
         columns.add(Scene.COLUMN_SCENE_KEY);
-        columns.add("ST_AsText(ST_Transform(" + Scene.COLUMN_FOOTPRINT + ",4326)) AS footprint");
+        columns.add("ST_AsText(ST_Transform(" + Scene.COLUMN_FOOTPRINT + ",4326),5) AS footprint");
```
danielkryska, line 18: What mean value `, 5` in this context?

danielkryska (approved):

s4e-backend/src/main/java/pl/cyfronet/s4e/data/repository/query/QueryBuilderImpl.java
jswk: It's the maximum numer of decimal digits, the same as I use in java format. I add this parameter here to keep the output generated by postgis (in search endpoints) and java (in scene endpoint) consistent.
See https://postgis.net/docs/ST_AsText.html for the postgis reference.

kmarszalek (approved):

PR #1049, opened 2020-12-23
Add an active scene polygon to the map

danielkryska:
On select a scene show polygon on map

image

### DIFF
--- a/s4e-web/src/app/views/map-view/map/map.component.ts
+++ b/s4e-web/src/app/views/map-view/map/map.component.ts
@@ -43,6 +43,7 @@ import {SentinelSearchService} from '../state/sentinel-search/sentinel-search.se
 import {SentinelSearchQuery} from '../state/sentinel-search/sentinel-search.query';
 import BaseLayer from 'ol/layer/Base';
 import {getPointResolution} from 'ol/proj';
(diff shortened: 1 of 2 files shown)


s4e-web/src/app/views/map-view/map/map.component.ts
```diff
@@ -302,6 +307,40 @@ export class MapComponent implements OnInit, OnDestroy {
     mapLayers.extend(activeLayers);
   }
 
+  private _replaceScenePolygon(scene: Scene | null) {
+    const hasPolygon = this.map.getLayers().getArray()
+      .includes(this._scenePolygon);
+    if (hasPolygon) {
+      this.map.removeLayer(this._scenePolygon);
+    }
```
jswk, line 315: You're calling this method in `updateLayers`, where you clear `mapLayers`, so I don't think you have to remove the polygon layer (it won't be on the layer list).
You probably also don't need the null-check below, as you check for this in `updateLayers`.
If you apply my suggestions I would also rename the method to `_setScenePolygon(scene: Scene)`.

jswk (approved):

kmarszalek (approved):

Issue #957, opened 2020-11-29
Problem with the correct search of the town after pressing enter


Marcelinna:
Tested 
 SOK version master-e9cfc9a443e542addb7453e720e1526aa263abc4

Steps:
1. Go to map/products?
2. Type "Wars" to Search and fast click ENTER

What happens?
Searched city is Abramowice Kościelne

What was expected to happen?
Searched city is Warszawa

kmarszalek:
Closed via #1016

PR #1056, opened 2020-12-23
Fix login backlink, fix locations search, fix redirect to settings dashboard

danielkryska:
- Fix fast selection of location in location search
- Fix active location color in location search
- Fix active location selection with mapa marker icon in location search
- Fix set/unset larger font
- Add map scale in map left top corner

### DIFF
--- a/s4e-web/src/app/components/root/root.component.ts
+++ b/s4e-web/src/app/components/root/root.component.ts
@@ -44,13 +44,16 @@ export class RootComponent implements OnInit, OnDestroy {
       .pipe(untilDestroyed(this))
       .subscribe(isHeightContrast => isHeightContrast
         ? this.document.body.classList.add('wcag_hc')
-        : this.document.body.classList.remove('wcag_hc'));
+        : this.document.body.classList.remove('wcag_hc')
+      );
 
+    const htmlTag = this.document.getElementsByTagName('html')[0];
     this.viewConfigurationQuery.select('largeFont')
       .pipe(untilDestroyed(this))
       .subscribe(isLargeFont => isLargeFont
-        ? this.document.body.classList.add('wcag_fs')
-        : this.document.body.classList.remove('wcag_fs'));
+        ? htmlTag.classList.add('wcag_fs')
+        : htmlTag.classList.remove('wcag_fs')
+      );
   }
 
   /**
--- a/s4e-web/src/app/components/search/search.component.html
+++ b/s4e-web/src/app/components/search/search.component.html
@@ -23,7 +23,7 @@
 >
   <section class="section search">
     <input
-      (keydown.enter)="select(query.getActive())"
+      (keydown.enter)="selectActive()"
       (keydown.arrowup)="activatePreviousResult()"
       (keydown.arrowdown)="activateNextResult()"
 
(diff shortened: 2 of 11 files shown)


jswk (changes requested):
The backlink handling on unauthorized seems to work (although it works on master instance too, so it's a tad confusing).

However, it's not possible to navigate the place-search results by arrows now. The current item isn't highlighted anymore. Moreover, the bug with selecting "Abramowice Kościelne" when inputting something else and hitting enter before the results are loaded remains, but I don't know if it was in the scope of the PR ("fix search location on map" is unclear to me).

Marcelinna:
as @jswk  mentioned
- lack of highlighting the city when navigating with arrows
- problem with the correct search of the town after pressing enter
(https://github.com/cyfronet-fid/sat4envi/issues/957)

Marcelinna:
lack of highlighting the city when navigating with arrows still exist

jswk (changes requested):
Tested on 581ba967.

#### Authorization problem
1. Go to map, unauthorized (logout if needed).
2. You get redirected to login, and cannot access the map. (If you click "przejdź do mapy", then you are redirected back to login).

#### Search navigation
1. Focus the search bar. Wait for basic results.
2. Type quickly "war" and hit Enter. Nothing happens, this problem is fixed.
3. When you hit enter again upon results loading you get to Wargawa.
4. Hit down arrow. Results show up. Try to navigate the results with arrows, go down 1 position to Wara, click Enter. Nothing happens, I would expect to be able to select an alternative result though.
5. This behaviour is inconsistent though. When you type Wargowo and are presented with two results, then choosing either one using the same method (with arrows) works. Yet, when you type "Tychów " and also two results appear, then choosing doesn't work again.

There's some inconsistency in this behaviour. It doesn't seem like blocker to me though. Nevertheless, not being able to browse as an unauthorized user requires a fix.

danielkryska:
> Tested on [581ba96](https://github.com/cyfronet-fid/sat4envi/commit/581ba967aaf33920d508f0da39c8c96ec00c35b3).
> 
> #### Authorization problem
> 1. Go to map, unauthorized (logout if needed).
> 2. You get redirected to login, and cannot access the map. (If you click "przejdź do mapy", then you are redirected back to login).
> 
> #### Search navigation
> 1. Focus the search bar. Wait for basic results.
> 2. Type quickly "war" and hit Enter. Nothing happens, this problem is fixed.
> 3. When you hit enter again upon results loading you get to Wargawa.
> 4. Hit down arrow. Results show up. Try to navigate the results with arrows, go down 1 position to Wara, click Enter. Nothing happens, I would expect to be able to select an alternative result though.
> 5. This behaviour is inconsistent though. When you type Wargowo and are presented with two results, then choosing either one using the same method (with arrows) works. Yet, when you type "Tychów " and also two results appear, then choosing doesn't work again.
> 
> There's some inconsistency in this behaviour. It doesn't seem like blocker to me though. Nevertheless, not being able to browse as an unauthorized user requires a fix.

Going to map as an unauthorized user have been fixed.

jswk (approved):
The unauthorized bug (see [Authorization problem in my previous comment](https://github.com/cyfronet-fid/sat4envi/pull/1056#pullrequestreview-561695799)) is now fixed. Apart from the non-blocking problems with search it's ok to merge.

Marcelinna:
please delete "fix redirect to settings dashboard" from name od PR, this bug isnt fixed. 
And in the description write  "Add map scale in map left top corner". Scal is added to bottom left corner.

Marcelinna (approved):

PR #1073, opened 2021-01-12
Cypress debugging, performance and memory leak patch

danielkryska:
- Run e2e in quiet mode without recording and photo making
- Change e2e engine to chrome in headless mode to get more performance options
- Set up browser one tab space and general shared space
- Display failed tests logs in terminal instead of making images or videos of failed test case


michal-szostak (approved):

Marcelinna (approved):

Marcelinna (approved):

jswk (changes requested):
If you're disabling screenshots, then remove the `E3E upload screenshots` step from `e2e.yml` workflow.

It also seems that the maxListeners is still too low, as there are warnings in [the logs](https://github.com/cyfronet-fid/sat4envi/pull/1073/checks?check_run_id=1694034105):
```
(node:7982) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 preprocessor:close listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
```
You should probably try to match the number of spec files, so it seems.

jswk (approved):

PR #1077, opened 2021-01-13
Update and run ignored e2e tests, add e2e tests for institution features, increase page objects granularity

Marcelinna:
Update and run ignored e2e tests
- update user profile change password to new layout/flow
- replace institution people management test with new one suit to new layout/flow
- replace institution profile management test with new one suit to new layout/flow
- update the class name to include the main test division: auth, map, setting, useroptions ( LocationsSearch -> MapLocationSearch)

Add e2e tests for institution features
- add institution delete privileges test
- add institution layers management test
- add institution products management test
- add institution.json file with the list of institution used in tests (add, edit, remove institution)
- add cat1user.json file with email and password data for the login used in tests( manage institution products)

Delete e2e tests (for non-existent functionalities or change distribution of tests)
- delete settings-breadcrumbs test
- delete settings-dashboard test
- delete settings-institutions test
- delete settings-invitations-list test
- settings-user-profile test

Changing the division of tests
- new division for user settings and administration panel test and page object model, more corresponding with the name of the feature in the application


s4e-web/cypress/integration/settings/settings-add-delete-edit-institution.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      .gotoUserProfile()
+    SettingsNav
+      .goToInstitutionList()
+    
+    AddInstitution.pageObject.getInstitutionCount();
+  });
+
+  it('should add, edit and delete new institution', function () {
+    AddInstitution
+      .openAddNewInstitutionPage()
```
danielkryska, line 47: goTo

s4e-web/cypress/page-objects/settings/settings-superAdmin-add-privilege.po.ts
```diff
@@ -0,0 +1,26 @@
...
+  static toggleDeleteInstitutionPrivilege(user: string, privilege: boolean) {
+
+    SettingsAdministratorPrivilage
+      .pageObject
+      .getAdministrator()
+      .contains(user)
+      .parent()
+      .find(SettingsAdministratorPrivilage.pageObject.getToogleDeleteInstitutionPrivilageClass)
+      .then(($element) => {
+        if ($element.is(':not(:checked)') && privilege) {
```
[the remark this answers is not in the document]
danielkryska: superAdmin in file name should be lowercase (`superadmin`)

s4e-web/cypress/integration/map/map-data-search.spec.ts
```diff
@@ -31,28 +31,34 @@ describe('Sentinel data search', () => {
 
   context("Data search for logged in user", () => {
 
-    beforeEach(function(){
+    beforeEach(function () {
```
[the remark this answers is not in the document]
Marcelinna: this gap it is in every test in beforeEach, this code formatting comes from VSC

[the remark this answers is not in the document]
jswk: `function(){` seems to be a typo, Marcelinna corrected it to `function () {`.

s4e-web/cypress/integration/settings/settings-change-password.spec.ts
```diff
@@ -1,30 +1,36 @@
...
+  it('should change password and login with it', function () {
+    UserOptionsGoToSettings
+      .gotoUserProfile();
+    SettingsUserPasswordChange
+      .goToChangePasswordPage()
+      .changePassword(this.zkAdmin.password, this.zkAdmin.password.toUpperCase());
+    SettingsNav
+      .logOut();
+    Login
+      .loginAs({ ...this.zkAdmin, password: this.zkAdmin.password.toUpperCase()});
```
jswk, line 29: I'd consider trying to use outdated credentials too, verifying that you get an error when logging in.

s4e-web/cypress/integration/settings/settings-institution-layer.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  cy.fixture('layer-capability.xml').as('geoserverResponse');
+});
+
+describe('Institution Layer', () => {
+
+  beforeEach(function () {
+    cy.visit('/login')
+    Login
+      .loginAs(this.zkAdmin)
+    MapLayers.pageObject.getLayerCount();
```
jswk, line 22: Why is such call necessary in the beforeEach?

s4e-web/cypress/integration/settings/settings-change-password.spec.ts
Marcelinna: you think about deleted account ?

jswk: No, I'm thinking about such course of action:
1. Login using password A (initial password value)
1. Change pass from A to B
2. Logout
2. Try to login using password A (should fail)
3. Login using password B (should succeed)

s4e-web/cypress/integration/settings/settings-institution-layer.spec.ts
Marcelinna: I check the number of layers in the application before the test. In the test, I add 1 layer to the institution and check if the number of layers has increased by 1

jswk: Logging-in is understood. Don't the layer count assertions from `map-layers.po.ts` call pageObject themselves? E.g. in the function you use to verify the count has risen by 1, it reads:
```ts
  static sidebarLayersCountShouldBe(count: number) {
    MapLayers
      .pageObject
      .getSidebarLayers() // there the pageObject is called
      .should('have.length', count);

    return MapLayers;
  }
```
I understand the call `.as('layerCount')` in `getLayerCount` assigns the value under this to placeholder (?) `layerCount`, but then there are no references to it throughout the code (it would read `@layerCount`, at least according to docs I have loaded).
image

s4e-web/cypress/integration/settings/settings-institution-profile.spec.ts
```diff
@@ -1,91 +1,138 @@
...
+import { UserOptionsAuthentication } from '../../page-objects/user-options/user-option-authentication.po';
 
-// context.skip('Settings institution profile', () => {
-//   beforeEach(() => {
-//     cy.fixture('users/zkAdmin.json').as('zkAdmin');
-//   });
+before(() => {
+  cy.fixture('users/zkAdmin.json').as('zkAdmin');
+  cy.fixture('users/admin.json').as('admin');
+  cy.fixture('institution.json').as('institution');
```
jswk, line 14: Shouldn't this fixture have plural name? I.e. `institutions` instead of a singluar `institution`.

jswk (approved):
Apart from the few non-blocking comments it seems fine.
I'd only insist on changing the fixture name to plural, as it is misleading with singular name.

s4e-web/cypress/integration/settings/settings-institution-layer.spec.ts
Marcelinna: .should('have.length', count) it is only assertion, this function doesnt allow use the count because doesnt return the number.

jswk: > .should('have.length', count) it is only assertion, this function doesnt allow use the count because doesnt return the number.

Okay, but it still doesn't explain why the call in the beforeEach is necessary. It seems to me that the line `    MapLayers.pageObject.getLayerCount();` can be safely removed, it's a no-op.

jswk: I'd suggest you to try removing it, see if the tests pass:
- if they pass, leave it removed
- if they fail, re-add it and give me a heads-up that it is necessary after all

s4e-web/cypress/integration/settings/settings-institution-profile.spec.ts
```diff
@@ -1,91 +1,138 @@
...
+import { UserOptionsAuthentication } from '../../page-objects/user-options/user-option-authentication.po';
 
-// context.skip('Settings institution profile', () => {
-//   beforeEach(() => {
-//     cy.fixture('users/zkAdmin.json').as('zkAdmin');
-//   });
+before(() => {
+  cy.fixture('users/zkAdmin.json').as('zkAdmin');
+  cy.fixture('users/admin.json').as('admin');
+  cy.fixture('institutions.json').as('institution');
```
jswk, line 14: I'd modify the `.as('institution')` to `.as('institutions')` too, as it's still misleading.

s4e-web/cypress/integration/settings/settings-change-password.spec.ts
jswk: Perfect, thanks!

s4e-web/cypress/integration/settings/settings-institution-layer.spec.ts
Marcelinna: I get every element for test in page object files but tests execute to the another file spec.ts. So I cant use directly cy.get(@layerCount) in spec.ts. I have to create another function in Page Object file for cy.get(@layerCount) and import and trigger to spec.js file where I have tests. I will think which solution will be better.

Marcelinna: it is necessary, without this test is failed 
image

jswk: Okay, I understand now. Sorry for the bother.

The pageObject.getLayerCount actually **sets** (and not gets) the observed layer count and puts it in `this.layerCount` (I didn't know you can access it this way) that you use for assertion later on. I'd rename it from `getLayerCount` to `setAliasLayerCount` for better clarity too.

danielkryska (approved):

Issue #1079, opened 2021-01-13
Group products in search panel based on metadata schemas


jswk:
Group products in search panel based on metadata schemas, as initially implemented in scope of #1040.

PR #1080, opened 2021-01-13
[#1079] Group products in search panel based on metadata schemas

jswk:
The products with undefined SectionPrototype are by default placed after
all defined sections. The extra sections are ordered by schema name.

Closes: #1079.


s4e-backend/src/main/java/pl/cyfronet/s4e/controller/ConfigController.java
```diff
@@ -67,10 +67,10 @@ public ConfigResponse config() {
     @ApiResponses({
             @ApiResponse(responseCode = "200", description = "OK")
```
danielkryska, line 68: I see that you're using user data inside of controller, but you don't included the 401, 403 errors in the Api responses. Does it have a purpose?

danielkryska (approved):

s4e-backend/src/main/java/pl/cyfronet/s4e/controller/ConfigController.java
jswk: Here omitting 401/403 is intentional, as these endpoints should always return 200.
Product filtering in the supplier only narrows down the list of returned products.

kmarszalek (approved):

Issue #1083, opened 2021-01-18
Product categories sorting


jswk:
Sort product categories according to their ranks in the product list, currently the sections are unsorted:
![image](https://user-images.githubusercontent.com/583197/104934338-4ca31280-59aa-11eb-806e-7e52188c76dd.png)

PR #1085, opened 2021-01-19
[#1083] Sort categories and products by rank

danielkryska:


michal-szostak (approved):

jswk (changes requested):
Changing product_category rank doesn't result in order change. See for example current state of https://pr-1085-sok.docker-fid.grid.cyf-kr.edu.pl.
I changed the rank of the product_category other to 10000, so it should be sorted last, instead of first. Yet it still persists on top.

jswk (changes requested):
The categories order is now ok,however, now the order of products in categories is reversed, see:
image
While the correct order is:
image

Add a test of this sorter to save our time when changing something around this functionality.
It should at least include three categories, and three products in at least one of them, to see if the sorting works fine. The products with categories should come in out-of-order, so that it's verified it's not a mistake it sorts them this way.

jswk (approved):

Issue #1086, opened 2021-01-19
[Fix] displaying the profile of a child-institution after it is added


Marcelinna:
Tested on instance SOK version master-f3e4922d1a4e179a091870c0777ac80b300fe5d5

**Steps:**
1. Login as zkAdmin
2. Go to /settings/institutions
3. Choose Zarządzenie kryzysowe - PL
4. Click on the button <_Dodaj jednostke podlegla>
5. Fill required field and "E-maile administratorów" field too 
6. Click on the button <_Zapisz instytucje> 

**What happens?**
![image](https://user-images.githubusercontent.com/59872096/105032980-cf35db80-5a57-11eb-9fda-8ff7a5057786.png)

**What was expected to happen?**
![image](https://user-images.githubusercontent.com/59872096/105033209-1b811b80-5a58-11eb-8a28-39f6abe7d30d.png)

Marcelinna:
works on master-6cec7fb3e009e641b4e0a3449b6233a0f0741692

PR #1089, opened 2021-01-21
[#1086] [Fix] update user privilege to a new institution, [Fix] update map navigation to settings

danielkryska:
- update update user privilege before sending invitations to the users on child-institution addition
- [Fix] inconsistent /settings redirection from /map


s4e-web/src/app/views/settings/settings.component.html
```diff
@@ -26,7 +26,11 @@
   <nav
     class="navigation"
   >
-    <a routerLink="/settings">
+    <a [routerLink]="
+      (hasAnyAdminInstitution$ | async)
+        ? (isAdminOfOneInstitution$ | async) ? '/settings/dashboard' : '/settings/institutions'
+        : '/settings'
+    ">
```
michal-szostak, line 33: This is a little to much nested logic in `routerLink` it might be better to make stream in component and only pipe it with `async` here.

michal-szostak (approved):

s4e-web/src/app/views/settings/settings.component.html
danielkryska: I was considering it, now I'm sure that I should do it from beginning :P

Marcelinna:
When I login in to zkAdmin (User with institution privilege) I see only "Twoj profil". When I am in /settings/institutions and come back in the map, in user options I see "Twoj profil" and "Panel administracyjny". Why right after logged in I don't see "Panel administracyjny"?

Marcelinna:
Please change in header [#1083] [Fix] update user privilege to a new institution **to** **[#1086]** [Fix] update user privilege to a new institution

Marcelinna (approved):

PR #1104, opened 2021-01-28
[#1103] Add EUMETSAT license if user already has a defined authority

jswk:
Add LICENSE_EUMETSAT to user's authorities if she already has an
authority from an eumetsat_authority_whitelist DB property.
In case no such property is defined then default to
ROLE_MEMBER_ZK and ROLE_MEMBER_PAK.

By using the property the role range can be narrowed down.

Closes: #1103.


kmarszalek (approved):
maybe some test?

danielkryska (approved):

PR #1114, opened 2021-02-08
[#1096] Add SyncRecordController to expose synchronization history

jswk:
Use QueryDSL to expose a flexible search endpoint.
Add an authority OP_SYNC_RECORD_READ to allow access to this endpoint,
it can be assigned to a user by an admin.

Admin can also delete records by an arbitrary search query.

Populate records in SyncRecorder.
To get all the necessary data I had to modify the Context to accommodate
extra data.

Closes: #1096.

### DIFF
--- a/s4e-backend/src/main/java/pl/cyfronet/s4e/sync/ContextRecorder.java
+++ b/s4e-backend/src/main/java/pl/cyfronet/s4e/sync/ContextRecorder.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2021 ACC Cyfronet AGH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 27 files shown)


s4e-backend/src/main/java/pl/cyfronet/s4e/sync/ContextRecorder.java
```diff
@@ -0,0 +1,71 @@
...
+            if (context.getProduct() != null) {
+                syncRecordBuilder.productName(context.getProduct().getName());
+            }
+            if (error != null) {
+                syncRecordBuilder.resultCode("err:" + error.getCode());
+                syncRecordBuilder.parameters(error.getParameters());
+                if (error.getCause() != null) {
+                    syncRecordBuilder.exceptionMessage(error.getCause().getMessage());
+                }
+            } else {
```
danielkryska, line 63: Is the else code block needed? Can it be flattened?

danielkryska (approved):

s4e-backend/src/main/java/pl/cyfronet/s4e/sync/ContextRecorder.java
jswk: Yes, it's needed, as otherwise every SceneRecord would have `resultCode="ok"`.

kmarszalek (approved):

PR #1120, opened 2021-02-10
Node angular upgrade

michal-szostak:
## What is in this PR?

Upgrade following dependencies

* Angular -> 11
* Node -> 14
* Cypress -> 6
* Typescript -> 4

WIP - as part of the upgrades owlDatepicker has became obsolete and incompatible - switched to ngx-bootstrap, but it requires adeditional styling


s4e-web/.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 14.15.5
```
jswk, line 1: There is already a .tool-versions in project root with nodejs version declared. I'd update it there.

michal-szostak: alright - thanks for pointing it out - this PR is still in 'disaster state'...
Should I watch out for any other npm / node version apart from `s4e-web/pom.xml` and that `.tool-versions` ?

jswk: > Should I watch out for any other npm / node version apart from `s4e-web/pom.xml` and that `.tool-versions` ?

A search through files only results in three places with nodejs version:
image
I believe npm version isn't specified anywhere, apart from README files, which could be also updated, but won't result in any failure.

michal-szostak: Ok, I've moved `.tool-versions` to the root folder and updated references to node / npm.

Marcelinna (approved):

abacz (approved):

PR #1122, opened 2021-02-11
[#1121] Expose helpdesk properties in configuration response

jswk:
The properties can be defined by setting `helpdesk_config` DB property
in the format `<key>=<value>(,<key>=<value>)+`.

In development seed its value in SeedHelpdeskConfig.

Fixes: #1121.


michal-szostak (approved):
Should work - I'm not entirely sure about the default seeds thought.

kmarszalek (approved):

PR #1127, opened 2021-02-12
Add product label name to the load view modal, fix infinite loading while fetching map product

danielkryska:
IMPORTANT!!! 
Bakground loader never ends due to issue linked with never triggered event renderend on map. Occurs in case of CORS.

- Force map loader to stop on
  - View change
  - Scene change
- Add product label name to the load view modal


Marcelinna:
I don't know that 5 sec is enough? on production, for example, for " Charakterystyka powierzchni o wysokiej rozdzielczości " after 5 sec nothing has loaded yet. I have to wait longer, and all scene is loading approximately after 30 seconds.

jswk:
> I don't know that 5 sec is enough? on production, for example, for " Charakterystyka powierzchni o wysokiej rozdzielczości " after 5 sec nothing has loaded yet. I have to wait longer, and all scene is loading approximately after 30 seconds.

I also think that 5 seconds isn't long enough, although I would rather not have any timeout on this. Having the loader reset on scene change (and therefore product change) is enough.
I think a user would try navigating to another scene anyway, if she encounters a problem with loading, thus resetting the loader.

Btw, one possibly extra case for loader reset: overlay change, as it triggers reload as well.

Marcelinna:
extra case for loader reset too: turning off the product.

danielkryska:
Timeout made to stop background after time of 5 sec have been removed

Marcelinna (approved):

jswk (approved):
Correct the format so that liner passes. Otherwise, the PR is fine now.

PR #1152, opened 2021-02-25
Improve API description

jswk:
Add security information to OpenAPI description and how-to.
Correct description of a few endpoints.
Group API endpoints and expose their definitions to selected authorities.
Add search endpoint query parameters to OpenAPI.

### DIFF
--- a/s4e-web/src/app/views/apihowto/apihowto.component.html
+++ b/s4e-web/src/app/views/apihowto/apihowto.component.html
@@ -24,7 +24,7 @@
     BŁĄD:<br />
     curl: (60) SSL certificate problem: unable to get local issuer certificate More
     details here: https://curl.haxx.se/docs/sslcerts.html<br />
-    Aby połączyć się z HOST w sposób niebezpieczny, można użyć `-k'.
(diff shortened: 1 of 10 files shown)


s4e-web/src/app/views/apihowto/apihowto.component.html
```diff
@@ -49,7 +51,7 @@ <h2>Pobranie klucza API (tzw. token)</h2>
 
   <label>Przykład:</label>
   <code>
-    curl -LX POST '{{ API_BASE }}/token' -H 'Content-Type: application/json'
+    curl -LX POST '{{ API_BASE }}/token' --header 'Content-Type: application/json'
```
kmarszalek, line 54: is there any difference between -H and --header? curl help doesn't mention anything :(

jswk: It's just more consistent, both in wget and curl you have --header, but only in curl there is -H which expands to --header (it's something completely different in wget).

danielkryska (approved):

kmarszalek (approved):

Issue #1182, opened 2021-04-16
The "Na zywo" closing modal is displayed incorrectly


Marcelinna:
Tested on https://dane.sat4envi.imgw.pl/
SOK version v17.6.0-RELEASE

1. Go to https://dane.sat4envi.imgw.pl/map/products?
2. Chose product (e.g Pyl w atmosferze)
3. Select a date from the calendar with no scene (e.g 2022-06-08)
4. Click button ""Na zywo"
5. Modal for closing is displayed

![image](https://user-images.githubusercontent.com/59872096/115001786-4555ad00-9ea4-11eb-9e03-fac48eaf9292.png)

Expected behaviour: Modal for closing "Na zywo" mode isnt displayed

PR #1203, opened 2021-07-21
[#1182] Fix modal display on live mode exit

danielkryska:


Marcelinna (approved):

michal-szostak (approved):

Marcelinna:
Please make sure you are not introducing regression. When you click on the "Live" button, the last available scene is not shown, but the first scene of the day

1. Mikrofizyka chmur
PR: 
image

Master:
image

2.Detekcja chmur lodowych i sniegu
PR: 
image

Master:
image

3. Chmury konwekcyjne
PR:
image

Master:
image

Marcelinna (dismissed):
blocked due to comment above

jswk (changes requested):
Okay, after having a look at the behaviour before and after, the regression is actually related with the fix.

On the master instance, if I select a product, go to not-the-last-with-scenes-available day (it can have scenes, it's only important that the app has to switch days to arrive at the last-with-scenes-available day) and toggle live-mode, then always a correct scene is selected, i.e. the last one on that day. The modal is simultaneously displayed, but it doesn't affect the result.

On the PR instance however, the behaviour is different, and the correct scene is selected for a moment, but then it gets switched to the first on that day.

This is indeed a regression, related with the fix. And as @Marcelinna correctly noted we shouldn't introduce regressions with fixes.
I understand, that you @danielkryska didn't change other places, but it seems that by removing the part of the logic from `selectDate`, i.e.:
```typescript
    const turnOfLiveMode = await this.timelineService.confirmTurningOfLiveMode();
    if (!turnOfLiveMode || date == null) {
      return;
    }
```
messed up the akita event processing order, introducing a regression.
Therefore, this has to be considered in scope of the same issue and corrected, so that you regain previous behaviour, which was correct.

I'm unable to fully track the logic there, as it's greatly convoluted, but my suspicion is on you having set `this._sceneStore.setLoading(true);` in `selectDate` before, and not doing it now.
You have this chunk of code depend on it:
```typescript
  private _handleUpdater$ = this._sceneQuery.selectLoading().pipe(
    filter(isLoading => !isLoading),
    map(() => this._sceneQuery.getValue().isLiveMode),
    tap(isLiveMode => (isLiveMode ? this._turnOnUpdater() : this._turnOffUpdater()))
  );
```
and it possibly doesn't fire in a correct order anymore. If you're unable to pinpoint the issue, I would suggest to isolate the issue: remove other not-fixing changes from the PR (btw, you didn't catch all incorrect wording, grep `OfLive` in the project to catch all) this includes the change in `set _currentDate` in TimelineComponent (it's also a change that seems suspicious to me - it may change the propagation range of the event and thus introduce the regression) and verify that the issue persists. Only then dig down in the event order debugging.

PR #1206, opened 2021-10-21
Backend dependencies update - 2021.10

jswk:


.github/workflows/e2e.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v1
         with:
-          java-version: 11.0.10
+          java-version: 11.0.12
```
kmarszalek, line 12: why not 11.0.13? :)

